PR #10, opened 2021-04-07
Fixing linting and adding a more strict linter

Skarlso:
### Description

This PR will fix the linter checker to be more precise and aggressive. But still allows some wiggle room as confidence of golinter is set to 1 which ignores things like, error shouldn't be a full sentence and kClient shouldn't start with a k.

### Checklist

### Imports

Import statements now follow the following structure:

- stdlib
- thridparty
- organisational third party
- package local

### DIFF
--- a/pkg/catalog/search.go
+++ b/pkg/catalog/search.go
@@ -16,8 +16,7 @@ type HTTPClient interface {
 
 var httpClient HTTPClient = http.DefaultClient
 
-// Search will returns profiles for a given catalog by `catalogURL` which contain the give
-// name defined by `profileName`.
+// Search will return profile descriptions for a given `catalogURL` and `profileName`.
 func Search(catalogURL, profileName string) ([]ProfileDescription, error) {
 	u, err := url.Parse(catalogURL)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


aclevername (approved):
LGTM :smile:

pkg/catalog/search.go
```diff
@@ -7,13 +7,16 @@ import (
...
+// handler code.
 //go:generate counterfeiter -o fakes/fake_http_client.go . HTTPClient
 type HTTPClient interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
 var httpClient HTTPClient = http.DefaultClient
 
+// Search will return profile descriptions for a given `catalogURL` and `profileName`.
 func Search(catalogURL, profileName string) ([]ProfileDescription, error) {
```
bigkevmcd, line 20: How about `"Search queries the catalog at catalogURL for profiles matching the provided profileName."` ?

(Personally, I think `profileName` isn't quite right here, because that implies that the client knows how the server works, but that's probably not an issue right now).

Skarlso: Yeah that sounds good to me.

Callisto13: I have changed this to be something like `query` in my upcoming PR 👍

Skarlso: Even better! :)

PR #16, opened 2021-04-16
Generate YAML files which the user can install in their own cluster

Skarlso:
### Description

Generate installable yaml files using a catalog and a profile name.

### Checklist

### TODO

Figuring out Ownership.


cmd/pctl/main.go
```diff
@@ -84,6 +88,65 @@ func showCmd() *cli.Command {
 	}
 }
 
+func installCmd() *cli.Command {
+	return &cli.Command{
+		Name:      "install",
+		Usage:     "generate configuration objects for later cluster setup",
```
aclevername, line 94: this doesn't seem right. Its more like "generate resources for installing a profile" or something along those lines

Skarlso: Yeah, I meant something along the way of saying, setup your cluster, like, your kube cluster, set it up with the profile you will be installing. :D It sounds weird, now reading it after the weekend. :D

cmd/pctl/main.go
```diff
@@ -114,3 +177,30 @@ func printProfile(profile profilesv1.ProfileDescription) error {
...
+		}
+		return nil
+	}
+	for i, a := range artifacts {
+		if err := generateOutput(i, a); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
aclevername, line 206: does this logic belong in `main`? Makes it harder to unit test

pkg/catalog/install.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}()
+
+	if resp.StatusCode == http.StatusNotFound {
+		return nil, fmt.Errorf("unable to find profile `%s` in catalog `%s`", profileName, catalogName)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("failed to fetch profile: status code %d", resp.StatusCode)
+	}
```
aclevername, line 46:
```suggestion
	if resp.StatusCode != http.StatusOK {
	        if resp.StatusCode != http.StatusOK {
		     return nil, fmt.Errorf("unable to find profile `%s` in catalog `%s`", profileName, catalogName)
		}
		return nil, fmt.Errorf("failed to fetch profile: status code %d", resp.StatusCode)
	}

```

aclevername, line 46:
```suggestion
	if resp.StatusCode != http.StatusOK{
	        if resp.StatusCode == http.StatusNotFound {
		     return nil, fmt.Errorf("unable to find profile `%s` in catalog `%s`", profileName, catalogName)
		}
		return nil, fmt.Errorf("failed to fetch profile: status code %d", resp.StatusCode)
	}
```

pkg/catalog/install_test.go
```diff
@@ -0,0 +1,167 @@
...
+	var (
+		fakeHTTPClient *fakes.FakeHTTPClient
+	)
+
+	BeforeEach(func() {
+		fakeHTTPClient = new(fakes.FakeHTTPClient)
+		catalog.SetHTTPClient(fakeHTTPClient)
+	})
+
+	When("there is an existing catalog and user calls install for a profile", func() {
```
aclevername, line 31: can you add some tests for the more common error paths :smile:

cmd/pctl/main.go
```diff
@@ -84,6 +88,65 @@ func showCmd() *cli.Command {
...
+			profilePath, catalogURL, err := parseArgs(c)
+			if err != nil {
+				_ = cli.ShowCommandHelp(c, "install")
+				return err
+			}
+
+			branch := c.String("branch")
+			subName := c.String("subscription-name")
+			namespace := c.String("namespace")
+			configValues := c.String("config-map-name")
```
aclevername, line 131: need to check these aren't empty? Or do we get that for free from the CLI package

Skarlso: They will have the default value.

pkg/catalog/install_test.go
Skarlso: Yep! 👍

cmd/pctl/main.go
Skarlso: Yeah, no. I need to extract this and write some nice tests for it somewhere. :) You're right.

Skarlso:
This will be re-written as per design decisions. So I'm closing this PR and will open another one with different functionality.

PR #18, opened 2021-04-20
Create a profile subscription to be installed into a cluster

Skarlso:
### Description

Issue https://github.com/weaveworks/profiles/issues/26

Keeping it simple, the git repository creation and Pull Request will be in a separate PR.

### Checklist


pkg/catalog/install.go
```diff
@@ -0,0 +1,112 @@
...
+
+// FileWriter is a Writer using a file as backing medium.
+type FileWriter struct {
+	Filename string
+}
+
+// Output writes the profile subscription yaml data into a given file.
+func (fw *FileWriter) Output(prof *profilesv1.ProfileSubscription) error {
+	e := kjson.NewSerializerWithOptions(kjson.DefaultMetaFactory, nil, nil, kjson.SerializerOptions{Yaml: true, Strict: true})
+	f, err := os.OpenFile(fw.Filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
Callisto13, line 29: might `0600` be a bit tight?

pkg/catalog/install.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// StdoutWriter is a Writer using stdout as backing medium.
+type StdoutWriter struct{}
+
+// Output outputs the yaml generated content for a profile to stdout.
+func (s *StdoutWriter) Output(prof *profilesv1.ProfileSubscription) error {
+	e := kjson.NewSerializerWithOptions(kjson.DefaultMetaFactory, nil, nil, kjson.SerializerOptions{Yaml: true, Strict: true})
+	return e.Encode(prof, os.Stdout)
+}
```
Callisto13, line 51: I don't know if we need the stdout writer. In the ticket description I wrote both examples as in "pick one we can live with for just now" 😀

pkg/catalog/install.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Install using the catalog at catalogURL and a profile matching the provided profileName generates all the
+// artifacts and outputs a single yaml file containing all artifacts that the profile would create.
```
Callisto13, line 54: just to confirm that this installation is just about installing a profile which can be found in a catalog? (ie we will deal with "random" installations of profiles not in a catalog later)

pkg/catalog/install_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+func (sw *StringWriter) Output(prof *profilesv1.ProfileSubscription) error {
+	e := kjson.NewSerializerWithOptions(kjson.DefaultMetaFactory, nil, nil, kjson.SerializerOptions{Yaml: true, Strict: true})
+	if err := e.Encode(prof, sw.output); err != nil {
+		return err
+	}
+	return nil
+}
+
+var _ = Describe("Install", func() {
```
Callisto13, line 32: can we have some failure case tests here as well please? 😄

Callisto13 (commented):
initial comments, looks good and simple overall.

are you adding integration tests here, or do you think there is no point since the next story will change all UX?

pkg/catalog/install.go
Skarlso: Ahhh. :D Super sorry, I thought this thing should be choosable:
> I should see a ProfileSubscription yaml generated (either in a file or stdout)

So I thought `or` here meant that it should be choosable. :) Anyways, the interface is useful in terms of testing.

pkg/catalog/install_test.go
Skarlso: Yes, ofc. :D I'm writing it now. :) Just was overeager with sharing. :D

pkg/catalog/install.go
Skarlso: Hmm, good point. I guess 0644 is good enough too.

Skarlso:
> are you adding integration tests here, or do you think there is no point since the next story will change all UX?

Unsure. 🤔 I better add them I guess. :) And see from there.

cmd/pctl/main.go
```diff
@@ -84,6 +84,71 @@ func showCmd() *cli.Command {
...
+func installCmd() *cli.Command {
+	return &cli.Command{
+		Name:      "install",
+		Usage:     "generate resources for installing a profile",
+		UsageText: "pctl --catalog-url <URL> install --subscription-name pctl-profile --namespace default --branch main --config-map-name configmap-name <CATALOG>/<PROFILE>",
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:        "subscription-name",
+				DefaultText: "pctl-profile",
+				Value:       "pctl-profile",
```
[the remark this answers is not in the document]
Skarlso: Good question. I thought something like, as fewer options I have to define to get things up and running, the better?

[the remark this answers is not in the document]
Callisto13: yeh i think that is fine

pkg/catalog/install.go
Callisto13: yeh the interface is cool, but i don't think we need the stdout just yet. we can add it back later if that changes

Skarlso:
@Callisto13 Even though I might update the integration test, at least I know what still should pass or the structure I'm working in. I think it was a good idea to add them. :) 👍

Callisto13:
Can you update the readme so I know what the command is in the demo 😁

pkg/catalog/install.go
```diff
@@ -0,0 +1,103 @@
...
+	}(f)
+	if err := e.Encode(prof, f); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Install using the catalog at catalogURL and a profile matching the provided profileName generates a profile subscription
+// writing it out with the provided profile subscription writer.
+func Install(catalogURL, catalogName, profileName, subName, namespace, branch, configMap string, writer Writer) error {
```
Callisto13, line 46: that is quite a lengthy signature, do we want a struct instead?

Skarlso: Yes, definitely. I can do that.

pkg/catalog/install.go
```diff
@@ -0,0 +1,103 @@
...
+	"os"
+
+	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
+	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
+)
+
+// Writer takes a profile and writes it out into a medium.
+type Writer interface {
```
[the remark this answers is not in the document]
Skarlso: I was thinking about that, indeed. But had the same thought process of refactoring when it's needed. :)

cmd/pctl/main.go
```diff
@@ -84,6 +87,79 @@ func showCmd() *cli.Command {
...
+
+			parts := strings.Split(profilePath, "/")
+			if len(parts) < 2 {
+				_ = cli.ShowCommandHelp(c, "install")
+				return errors.New("both catalog name and profile name must be provided")
+			}
+			catalogName, profileName := parts[0], parts[1]
+
+			fmt.Printf("generating subscription for profile %s/%s:\n\n", catalogName, profileName)
+			var w writer.Writer = &writer.FileWriter{Filename: filename}
```
Callisto13, line 147:
```suggestion
			w := &writer.FileWriter{Filename: filename}
```
for continuity?

suppose you could also inline this on 156, but that is a personal style choice so have no strong feelings

pkg/catalog/install.go
```diff
@@ -0,0 +1,88 @@
...
+	ConfigMap   string
+	Namespace   string
+	ProfileName string
+	SubName     string
+	Writer      writer.Writer
+}
+
+// Install using the catalog at catalogURL and a profile matching the provided profileName generates a profile subscription
+// writing it out with the provided profile subscription writer.
+func Install(cfg InstallConfig) error {
```
Callisto13, line 31: lovely lovely

cmd/pctl/main.go
Skarlso: Ah yes, I used the interface definition there before because of the possible overwrite after. This was previously an stdout write which was optionally overwritten but the variable had to stay an interface type. :)

Now, that's not needed, so yeah, I can remove it. :)

Callisto13 (approved):
lgtm 🚀

PR #31, opened 2021-04-27
Fixing the integration test part 2

Skarlso:
### Description

~~Looks like git is not installed on the image so it wasn't able to push.~~
Dang.. it looks like the token doesn't have enough permissions.

### Checklist

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -1,9 +1,7 @@
 name: Integration Test
 
 on:
-  push:
-    branches:
-      - main
+  pull_request: {}
 
 jobs:
   test:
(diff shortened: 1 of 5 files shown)


.github/workflows/integration_test.yaml
```diff
@@ -1,9 +1,7 @@
 name: Integration Test
 
 on:
-  push:
-    branches:
-      - main
+  pull_request: {}
```
Skarlso, line 4: I'll revert this once I know that the action works.

Skarlso:
Fix will come in this format: https://github.com/weaveworks/corp/issues/1659

Skarlso:
Opening a new one with latest stuff.

Issue #5, opened 2021-03-31
Command to install Profiles controller and CRD


Callisto13:
```sh
pctl install/bootstrap/prepare/seed # whatever
```

should install ProfileSub, ProfileCatalogSource and whatever else in my cluster

aclevername:
just the CRDs doesn't seem that useful, the whole `profiles` controller deployment + CRDs sounds useful, mirroring `flux bootstrap`

Skarlso:
Further things for prepare.....

- [ ] we need to give the manifests a namepsace. they can’t be in system So that’s a default I’ll add to all kustomize files. ( flux also does this )
- [ ] add missing rbac to manifests.tar.gz
- [ ] create a latest tag for the profiles-controller so the manifest installation can work right off the bat without the need to give a version
- [ ] deal with the kube-rbac-proxy sidecar which is created by our default kustomize build

Callisto13:
@Skarlso is this ticket closable or are there still bits to do?

PR #30, opened 2021-04-27
Add support for listing and getting installed profiles

aclevername:
### Description

```
./pctl list
NAMESPACE       NAME                    READY
default         nginx-profile-test      False
```

```
./pctl get --namespace default --name nginx-profile-test
Subscription    nginx-profile-test
Namespace       default
Ready           False
Reason          error when reconciling profile artifacts
 ```
The get output is pretty limited until https://github.com/weaveworks/profiles/issues/87 happens

### Checklist
- [ ] Added tests that cover your change
- [ ] Added/modified documentation as required (such as the `README.md`)
- [ ] Made sure the title of the PR is a good description that can go into the release notes
- [ ] Added a `kind` label to the PR (e.g. `kind/feature`)

PR #36, opened 2021-04-28
Add prepare command to prepare an environment to run profiles by installing all necessary components.

Skarlso:
### Description

https://github.com/weaveworks/pctl/issues/5

### Checklist


Skarlso:
TODO: Once releases work, write integration tests.

pkg/catalog/prepare.go
```diff
@@ -0,0 +1,132 @@
+package catalog
```
aclevername, line 1: does this belong in the catalog package? props worth its own package

Skarlso: Yeah, thought about that. It might warrant its own package. But I thought every command just sort of goes there. :D I can refactor. :)

Skarlso:
@Callisto13 Nope, those are part of the unit test thingy... I haven't planned on removing it since the test transport roundtrip function returns that as body.

pkg/catalog/prepare_test.go
```diff
@@ -0,0 +1,215 @@
...
+type mockTransport struct {
+	res *http.Response
+}
+
+func (t *mockTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	return t.res, nil
+}
+
+var _ = Describe("prepare", func() {
+	It("can prepare the environment with everything that profiles needs with dry run", func() {
```
Callisto13, line 26: can this `It` be a little clearer? like "when dry run is set, prepares all the manifests that profiles requires but does not apply them to the cluster" or something

pkg/catalog/prepare_test.go
```diff
@@ -0,0 +1,215 @@
...
+		Expect(err).NotTo(HaveOccurred())
+		files, err := ioutil.ReadDir(tmp)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(len(files) > 0).To(BeTrue())
+		Expect(runner.RunCallCount()).To(Equal(1)) // there should be no call because of dryrun.
+		arg, args := runner.RunArgsForCall(0)
+		Expect(arg).To(Equal("kubectl"))
+		Expect(args).To(Equal([]string{"apply", "-f", tmp, "--dry-run=client"}))
+	})
+	It("can prepare the environment with everything that profiles needs", func() {
```
Callisto13, line 62:
```suggestion
	It("prepares the environment with everything that profiles needs", func() {
```

Callisto13:
>@Callisto13 Nope, those are part of the unit test thingy... I haven't planned on removing it since the test transport roundtrip function returns that as body.

so when i run `pctl prepare` i end up with manifests in my dir? (or wherever i set location to)

Skarlso:
@Callisto13 Oh sh*t sorry, I completely misunderstood what you said. Yes, I will clean up the downloaded manifests.

Callisto13:
I think flux has something like `flux install --export > flux-system.yaml` for if people want that. but yeh let's default to cleaning for now and we can add that opt later.

can we also make the default location `os.TempDir`?

pkg/catalog/prepare_test.go
Skarlso: @Callisto13 Is it better now? I wrapped it into a When to make the context more clearer.

Callisto13: yep way nicer to read, ty!

README.md
```diff
@@ -22,6 +22,21 @@ generating subscription for profile nginx-catalog/weaveworks-nginx:
...
+
+```
+pctl prepare
+```
+
+This will take the latest manifests release under the profiles repository and install
+them into the currently set cluster.
+
+There are a number of options which can be set, such as: version, dry-run, context, kube-config.
+Please run `pctl help` for all options and defaults.
```
Callisto13, line 38:
```suggestion
Please run `pctl --help` for all options and defaults.
```
sorry i suggested the wrong thing for this 🤦‍♀️

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,132 @@
...
+		kubectlArgs = append(kubectlArgs, "--context="+kubeContext)
+	}
+	if kubeConfig != "" {
+		kubectlArgs = append(kubectlArgs, "--kubeconfig="+kubeConfig)
+	}
+	if _, err := a.Runner.Run(kubectlCmd, kubectlArgs...); err != nil {
+		return fmt.Errorf("install failed: %w", err)
+	}
+
+	if dryRun {
```
Callisto13, line 124: could you describe the UX of using dry run (beyond the obvious of not having stuff done to the cluster)? i am having trouble picturing it this early in the day😅 .

what outputs (stdout, files, etc) should i expect to see?

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,132 @@
...
+
+// Prepare will prepare an environment with everything that is needed to run profiles.
+func (p *Preparer) Prepare() error {
+	if err := p.Fetcher.Fetch(context.Background(), p.BaseURL, p.Version, p.Location); err != nil {
+		return err
+	}
+	if err := p.Applier.Apply(p.Location, p.KubeContext, p.KubeConfig, p.DryRun); err != nil {
+		return err
+	}
+	return nil
```
Callisto13, line 72: can we `return p.Applier.Apply(...` here?

pkg/cluster/prepare_test.go
```diff
@@ -0,0 +1,219 @@
...
+			}
+			err = p.Prepare()
+			Expect(err).NotTo(HaveOccurred())
+			files, err := ioutil.ReadDir(tmp)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(files) > 0).To(BeTrue())
+			Expect(runner.RunCallCount()).To(Equal(1)) // there should be no call because of dryrun.
+			arg, args := runner.RunArgsForCall(0)
+			Expect(arg).To(Equal("kubectl"))
+			Expect(args).To(Equal([]string{"apply", "-f", tmp, "--dry-run=client"}))
```
Callisto13, line 61: can we add a line to verify that the temp dir (Location) is cleaned up afterwards?

pkg/cluster/prepare_test.go
```diff
@@ -0,0 +1,219 @@
...
+				},
+				Applier: &cluster.Applier{
+					Runner: runner,
+				},
+			}
+			err = p.Prepare()
+			Expect(err).To(MatchError("failed to download manifests.tar.gz from https://github.com/weaveworks/profiles/releases/latest/download/manifests.tar.gz, status: Not Found"))
+		})
+	})
+	When("there is an error accessing the output folder", func() {
```
Callisto13, line 131: speaking of this, do we want to do a stat (or whatever) on this folder immediately so that we fail fast before any client action?

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,132 @@
...
+	if cfg.BaseURL == "" {
+		cfg.BaseURL = releaseUrl
+	}
+	return &Preparer{
+		PrepConfig: cfg,
+		Fetcher: &Fetcher{
+			Client: http.DefaultClient,
+		},
+		Applier: &Applier{
+			Runner: &git.CLIRunner{},
```
Callisto13, line 59: do we want to move this runner to its own lil package, rather than importing the git one (bit misleading)?

pkg/cluster/prepare.go
Skarlso: You're right. Right now, there is no output in case of dry-run. So I guess the UX is not really helpful, eh?

Callisto13: maybe print the manifests which would be applied to stdout? 🤷‍♀️

Skarlso: Ironically, you can always just view them if you download the release yourself? Maybe there is no need for dry-run. It's not like we are building them through Kustomize right now. We just download a release. Maybe if we would build it ourself there would be value in displaying them.

Skarlso: Maybe it's still good ux though... you can verify which release it did download. That makes sense. :)

Skarlso: > dry-run |   | none | Must be "none", "server", or "client". If client strategy, only print the object that would be sent, without sending it. If server strategy, submit server-side request without persisting the resource.
-- | -- | -- | --

kubectl --dry-run=client will already output stuff that would have been applied, so there is nothing to do from pctl's point of view.

Skarlso: Note, this works if I apply `--output=yaml` for example. Then it will also display yaml data in a list Kind with everything it would have applied.

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+	if err := p.Fetcher.Fetch(context.Background(), p.BaseURL, p.Version, p.Location); err != nil {
+		return err
+	}
+	return p.Applier.Apply(p.Location, p.KubeContext, p.KubeConfig, p.DryRun)
+}
+
+// Fetch the latest or a version of the released manifest files for profiles.
+func (f *Fetcher) Fetch(ctx context.Context, url, version, dir string) error {
+	ghURL := fmt.Sprintf("%s/latest/download/manifests.tar.gz", url)
```
Skarlso, line 80: Even though we allow a base url change so to test forks, this has to match though... And I'm okay with that part. Later we can add doing this with manual resources or something if downloading these manifest files is not something that can be done in an isolated environment from the network.

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,130 @@
...
+			Runner: &runner.CLIRunner{},
+		},
+	}, nil
+}
+
+// Prepare will prepare an environment with everything that is needed to run profiles.
+func (p *Preparer) Prepare() error {
+	if _, err := os.Stat(p.Location); err != nil {
+		return err
+	}
```
Callisto13, line 71: _possibly_ not necessary now, since it is no longer configurable, but 🤷‍♀️

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,130 @@
...
+	if kubeContext != "" {
+		kubectlArgs = append(kubectlArgs, "--context="+kubeContext)
+	}
+	if kubeConfig != "" {
+		kubectlArgs = append(kubectlArgs, "--kubeconfig="+kubeConfig)
+	}
+	if _, err := a.Runner.Run(kubectlCmd, kubectlArgs...); err != nil {
+		return fmt.Errorf("install failed: %w", err)
+	}
+
```
Callisto13, line 126: did we decide against clearing the temp folder?

pkg/cluster/prepare.go
Skarlso: We wanna make sure temp is accessible, yes?

pkg/cluster/prepare.go
Skarlso: Ah, right!

pkg/cluster/prepare.go
Callisto13: wouldn't 51 have failed if it wasn't tho?

Callisto13: meh, either way extra security is nbd

Skarlso: Possibly it would have.. 🤔

Callisto13: yeh it looks like `ioutil.Tempdir` does this:
```go
 try := filepath.Join(dir, prefix+nextRandom()+suffix)
 err = os.Mkdir(try, 0700)
 if os.IsExist(err) {
```
dir being os.TempDir in this case

Skarlso: :D Excellent. :)

pkg/cluster/prepare.go
Skarlso: Added a defer remove at the call location.

Callisto13 (approved):
:shippit-squirrel:

Skarlso:
> :shippit-squirrel:

:shipit:

Callisto13:
oops sorry about the merge conflicts 😆

Skarlso:
Yeah, no worries, I didn't merge this yet, because I want to create integration tests once I get the release running. :) Which is almost there. :D

README.md
```diff
@@ -55,6 +55,21 @@ generating subscription for profile nginx-catalog/weaveworks-nginx:
 
 Then the result will be in profile-subscription.yaml file.
 
+### Prepare
```
Callisto13, line 58: don't forget to run `mdtoc -inplace README.md` to update the TOC (i left instructions for what to download and run at the top of the readme)

Skarlso: Good call! :) I would totally have forgotten it.

Callisto13: worth adding that to a `make` something or other? 🤔

Callisto13: actually yeh, i will do that in another pr real quick: makefile needs a general tidy

Skarlso: I think so, yes. Anything that has to be run by hand will eventually be forgotten at least once, right? :) So making it as easy as possible is a good point.

I can add that in my PR?

Skarlso: Sounds good! :)

tests/integration/integration_test.go
```diff
@@ -301,6 +303,52 @@ status: {}
...
+
+	Context("prepare", func() {
+		When("prepare is called without any arguments", func() {
+			It("downloads and installs the latest release of manifest files from the profile repository", func() {
+				cmd := exec.Command(binaryPath, "prepare")
+				output, err := cmd.CombinedOutput()
+				Expect(err).ToNot(HaveOccurred())
+				Expect(string(output)).To(Equal("install finished\n"))
+			})
+		})
```
Callisto13, line 315: do we want to check that things are actually installed in the cluster?

Skarlso: Sure, although to be fair, this will change once there is a wait for ready state implemented in the prepare command.

Callisto13: hmmm okay i can wait for that, so long as we eventually test more than a log line 😉

Skarlso: Oh, I'm am writing a check, just wanted to know for the future. :D

cmd/pctl/main.go
```diff
@@ -282,6 +290,54 @@ func createPullRequest(c *cli.Context) error {
...
+				Keep:        c.Bool("keep"),
+			})
+			if err != nil {
+				return err
+			}
+			return p.Prepare()
+		},
+	}
+}
+
```
aclevername, line 340: can this be moved into a `prepare.go` in `cmd/pctl`? I've done that refactor for other commands in this PR, just to prevent one of us getting annoying merge conflicts :smile: https://github.com/weaveworks/pctl/pull/30

pkg/cluster/prepare_test.go
```diff
@@ -0,0 +1,325 @@
...
+				Applier: &cluster.Applier{
+					Runner: runner,
+				},
+			}
+			err = p.Prepare()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(runner.RunCallCount()).To(Equal(1))
+			arg, args := runner.RunArgsForCall(0)
+			Expect(arg).To(Equal("kubectl"))
+			Expect(args).To(Equal([]string{"apply", "-f", filepath.Join(tmp, "prepare.yaml"), "--dry-run=client", "--output=yaml"}))
```
aclevername, line 61: worth testing that `prepare.yaml` contains the correct `contents`?

aclevername (approved):
LGTM! couple comments, lovely tests

pkg/cluster/prepare.go
```diff
@@ -0,0 +1,158 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read body of the response: %w", err)
+	}
+
+	// make sure the right version of the controller is pulled
+	if hasVersionPrefix {
+		content = bytes.ReplaceAll(content,
+			[]byte("weaveworks/profiles-controller:latest"),
+			[]byte("weaveworks/profiles-controller:"+version))
+	}
```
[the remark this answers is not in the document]
Skarlso: That's a good question and brings me to even better idea wish I would have thought about before. :D
Right now, no. Sadly, the kustomize build isn't using versions. So every build will have `latest` as a tag. Which is not very nice, is it? So I think we should fix that.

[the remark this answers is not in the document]
Skarlso: Once this lands https://github.com/weaveworks/profiles/pull/105 this will no longer be needed since it will always point to the right version.

cmd/pctl/main.go
Skarlso: Yes! I actually wanted to do that. :D I saw you rearranged things.

pkg/cluster/prepare_test.go
Skarlso: It won't match since kube already applied some things like status, variables, a bunch of things that aren't in the original. So it's not like an `Equal` and that's that. I can compare individual things if I create a map yaml out of it... But that felt like overkill. :) Especially considering that we are testing the apply eventually in the integration tests.

tests/integration/prepare_test_cluster.go
```diff
@@ -0,0 +1,106 @@
...
+	cmd := exec.Command(binaryPath, "prepare", "--dry-run")
+	output, err := cmd.CombinedOutput()
+	if err != nil || !bytes.Contains(output, []byte("kind: List")) {
+		fmt.Println("Output of prepare was: ", string(output))
+		return fmt.Errorf("failed to run prepare command: %w", err)
+	}
+	fmt.Println("Install file generated successfully.")
+
+	fmt.Print("Replacing controller image to localhost:5000...")
+	output = bytes.ReplaceAll(output, []byte("image: weaveworks/profiles-controller:latest"), []byte("image: localhost:5000/profiles-controller:latest"))
```
Skarlso, line 51: @Skarlso Note to self, I need to replace this with a regex matcher, since this will now no longer be `latest` after https://github.com/weaveworks/profiles/pull/105 is merged.

pkg/cluster/prepare.go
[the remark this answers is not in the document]
aclevername: :+1:

PR #40, opened 2021-04-30
Update readme when 'make test'

Callisto13:
Also little makefile tidy

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 ##@ Test
 
-test: lint unit integration ## Run all tests
+test: lint unit integration docs ## Lint, run all tests and update the docs
 
 unit: ## Run the unit tests
 	ginkgo -r ./pkg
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,28 +1,36 @@
 ##@ Test
 
-test: lint unit integration ## Run all tests
+test: lint unit integration docs ## Lint, run all tests and update the docs
```
Skarlso, line 3: I feel like `test` shouldn't update docs. It's running all tests. That's it. Then again, it also runs lint, but I think I can accept lint because it's a code syntax test, right? But docs.... that generates potential output. :)

Callisto13: i see your point, but if it is a separate command then people will forget to run it, so we are back to square 1.

we do things like this in eksctl (regenerate all the things as part of test) otherwise there is no way to guarantee people will actually do it

Skarlso: BTW, I understand why you'd put it there, more likely that people don't forget about it, since you run make test before committing anyways... So I guess it's fine. :D

Skarlso (approved):
:shipit:

PR #48, opened 2021-05-12
Adding ability to install and show profiles via catalog name and versions

Skarlso:
### Description

https://github.com/weaveworks/profiles/issues/87

profiles PR: https://github.com/weaveworks/profiles/pull/112

Note, this will fail until the above PR is merged and pctl is updated to use the latest profiles.

### Checklist


pkg/catalog/show.go
```diff
@@ -4,20 +4,64 @@ import (
...
+}
+
 // Show queries the catalog at catalogURL for a profile matching the provided profileName
-func Show(catalogClient CatalogClient, catalogName, profileName string) (profilesv1.ProfileDescription, error) {
-	data, code, err := catalogClient.DoRequest(fmt.Sprintf("/profiles/%s/%s", catalogName, profileName), nil)
+func Show(catalogClient CatalogClient, opts ...ShowOpts) (profilesv1.ProfileDescription, error) {
+	options := &ShowOptions{}
+	for _, o := range opts {
+		o(options)
+	}
```
aclevername, line 49: How come you decided on the `opts` pattern? I think its slightly confusing because `profileName` and `catalogName` are mandatory, which doesn't really get conveyed with this pattern

cmd/pctl/show.go
```diff
@@ -39,8 +40,11 @@ func showCmd() *cli.Command {
 				return errors.New("both catalog name and profile name must be provided")
 			}
 			catalogName, profileName := parts[0], parts[1]
-
-			profile, err := catalog.Show(catalogClient, catalogName, profileName)
+			var catalogVersion string
+			if len(parts) == 3 {
+				catalogVersion = parts[2]
+			}
+			profile, err := catalog.Show(catalogClient, catalog.WithCatalogName(catalogName), catalog.WithProfileName(profileName), catalog.WithCatalogVersion(catalogVersion))
```
aclevername, line 47: we should enforce that version is provided. Nice feature down the line might be to make it optional and let pctl install the latest

pkg/catalog/show.go
Skarlso: funny that you ask. :D :D :D It's a nice story. So, I started with defining version and initially, only install cared about it, and it was strange adding it to show and then mostly I called it with `""` because show wasn't using it actually from anywhere else. But then I realised that now show can also use it, so I started defining it. And previously I didn't use url.Join in this nice way either. :D so I was checking if version is empty, which was ugly as well. :DDDD

ANYWAYS. Now I think it's obsolete too and was thinking about removing it, but wanted to know what you think about it. :D

cmd/pctl/show.go
Skarlso: so it should be required?

cmd/pctl/show.go
```diff
@@ -39,8 +40,11 @@ func showCmd() *cli.Command {
 				return errors.New("both catalog name and profile name must be provided")
 			}
 			catalogName, profileName := parts[0], parts[1]
-
-			profile, err := catalog.Show(catalogClient, catalogName, profileName)
+			var catalogVersion string
+			if len(parts) == 3 {
+				catalogVersion = parts[2]
+			}
+			profile, err := catalog.Show(catalogClient, catalogName, profileName, catalogVersion)
```
aclevername, line 47: do we want to error here if version isn't provided?

aclevername (approved):

Issue #33, opened 2021-04-28
pctl prepare should wait for the profiles controller to report ready before exiting


aclevername:
pctl prepare should wait for the profiles controller to report ready before exiting

see what flux does, or ask @Skarlso

PR #49, opened 2021-05-12
Wait for resources to be ready

Skarlso:
### Description

https://github.com/weaveworks/pctl/issues/33

### Checklist

### TODO

- [x] Make namespace configurable and default to `profiles-system`.
- [x] Somehow test the wait. I suspect that we won't... Since waiting can't happen on the prepare step because we are patching the image.


pkg/cluster/prepare.go
```diff
@@ -144,7 +155,10 @@ func (a *Applier) Apply(folder string, kubeContext string, kubeConfig string, dr
 		fmt.Print(string(output))
 		return nil
 	}
-	// In a follow up ticket, make this wait for all the possible resources to be condition=available.
-	fmt.Println("install finished")
+	fmt.Print("Waiting for resources to be verified...")
```
aclevername, line 158:
```suggestion
	fmt.Print("Waiting for resources to be ready...")
```

consistent with the error message below

pkg/cluster/prepare_test.go
```diff
@@ -94,11 +100,15 @@ var _ = Describe("prepare", func() {
...
+			Expect(waiter.WaitCallCount()).To(Equal(1))
+			args = waiter.WaitArgsForCall(0)
+			Expect(args).To(Equal([]string{"profiles-controller-manager"}))
 		})
 	})
 	When("context and config is provided", func() {
 		It("passes that over to kubernetes", func() {
-			runner := &fakes.FakeRunner{}
+			runner := &runnerfake.FakeRunner{}
+			waiter := &fakes.FakeWaiter{}
```
aclevername, line 111: could this be in a `BeforeEach` rather than re-defining reach time?

aclevername (approved):

Issue #32, opened 2021-04-28
pctl prepare should check if flux is installed


aclevername:
Flux is a requirement for profiles, lets inform the user if they try to install the profiles controller without it.

Skarlso:
Should we also check if flux is actually condition=available?

Callisto13:
document required set of permissions to install crds and check whether they are there

PR #50, opened 2021-05-13
Adding prepare flux checking

Skarlso:
### Description

https://github.com/weaveworks/pctl/issues/32

### Checklist


cmd/pctl/get.go
```diff
@@ -22,7 +22,7 @@ func getCmd() *cli.Command {
 				Name:        "namespace",
 				DefaultText: "default",
 				Value:       "default",
-				Usage:       "The namespace the subscrption is in",
+				Usage:       "The namespace the subscription is in",
```
aclevername, line 25: :joy:

pkg/cluster/prepare.go
```diff
@@ -78,12 +94,38 @@ func (p *Preparer) Prepare() error {
...
+		if output, err := p.Runner.Run(kubectlCmd, args...); err != nil {
+			fmt.Println("\nOutput from kubectl command: ", string(output))
+			return fmt.Errorf("failed to get crd %s : %w", crd, err)
+		}
+	}
+
+	fmt.Println("done.")
+	return nil
+}
+
```
aclevername, line 128: do we want to error if flux isn't installed, or just log a warning/hint to the user? I think it would be nice if we didn't introduce a dependency ordering for somebody setting a cluster up. Letting the user know they've installed profiles but it won't work until flux is installed might be a better UX. What do you think?

Skarlso: Agreed. We can output a warning on it. But warnings are a bit devious that they tend to be ignored and then the people have no idea what happened? Maybe we could add a flag to explicit bypass the pre-flight check error? Or something like... `--no-error-on-preflight` or something like that?

aclevername: > Maybe we could add a flag to explicit bypass the pre-flight check error? Or something like... `--no-error-on-preflight` or something like that?

That sounds like a reasonable middle ground :smile:

Skarlso: @aclevername Done. WDYT?

pkg/cluster/prepare.go
```diff
@@ -109,7 +110,11 @@ func (p *Preparer) PreFlightCheck() error {
 	args := []string{"get", "namespace", p.FluxNamespace, "--output", "name"}
 	if output, err := p.Runner.Run(kubectlCmd, args...); err != nil {
 		fmt.Println("\nOutput from kubectl command: ", string(output))
-		return fmt.Errorf("failed to get flux namespace: %w", err)
+		if p.IgnorePreflightErrors {
+			fmt.Println("WARNING: failed to get flux namespace. Flux is required for profiles to work.")
+		} else {
+			return fmt.Errorf("failed to get flux namespace: %w", err)
```
aclevername, line 116: shall` we print a hint about using this flag? e.g. 
```
failed to get flux namespace: <error>
if you wish to proceed with the installation without flux/pre-flight-checks please see the  --ignore-preflight-checks
```

Skarlso: Yeah that makes sense

Skarlso: Added.

pkg/cluster/prepare.go
```diff
@@ -20,6 +20,16 @@ const (
...
 )
 
+// FluxCRDs are CRDs which prepare is checking if they are present in the cluster or not.
+var FluxCRDs = []string{
+	"buckets.source.toolkit.fluxcd.io",
+	"gitrepositories.source.toolkit.fluxcd.io",
+	"helmcharts.source.toolkit.fluxcd.io",
+	"helmreleases.helm.toolkit.fluxcd.io",
+	"helmrepositories.source.toolkit.fluxcd.io",
+	"kustomizations.kustomize.toolkit.fluxcd.io",
```
aclevername, line 30: is this discover-able by referencing there respective packages? E.g. https://github.com/weaveworks/profiles/blob/9f749b8ceb974e4b51471d29412e8437b90300c3/pkg/profile/helm.go#L49-L50

Skarlso: Oh cool! Maybe. I'll check it out!

Skarlso: Done :D

aclevername (approved):

PR #54, opened 2021-05-17
pctl install generates the artifacts locally

aclevername:
### Description
Artifacts are generated locally. Example:

```
$ pctl install nginx-catalog/weaveworks-nginx/v0.1.0
generating subscription for profile nginx-catalog/weaveworks-nginx:

$ tree weaveworks-nginx
weaveworks-nginx
├── GitRepository-0.yaml
├── GitRepository-1.yaml
├── HelmRelease-2.yaml
├── HelmRelease-4.yaml
├── HelmRepository-5.yaml
├── Kustomization-3.yaml
└── profile.yaml
```

Currently the `profile.yaml` is the subscription, in order to preserve the `get/list` functionality. This can be changed in a future PR. I have also not addressed the failing `get/list` tests, that can also be fixed in a follow up PR. This PR is big enough :smile: 
### Checklist


cmd/pctl/install.go
```diff
@@ -43,12 +42,6 @@ func installCmd() *cli.Command {
 				Value: "",
 				Usage: "The name of the ConfigMap which contains values for this profile.",
 			},
-			&cli.StringFlag{
```
Skarlso, line 46: We should keep the ability to specify a location though.

aclevername: yeah thats fair. I `rm`'d this since its now a folder rather than a file. I'll add in a follow on PR. Keen to keep this as slim as possible :smile:

pkg/profile/helm.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (p *Profile) makeHelmRepoName(name string) string {
+	repoParts := strings.Split(p.subscription.Spec.ProfileURL, "/")
+	repoName := repoParts[len(repoParts)-1]
+	return join(p.subscription.Name, repoName, name)
+}
+
+func (p *Profile) makeHelmRelease(artifact profilesv1.Artifact, repoPath string) *helmv2.HelmRelease {
+	var helmChartSpec helmv2.HelmChartTemplateSpec
+	if artifact.Path != "" {
```
Skarlso, line 37: On or the other **must** be defined, right..?

pkg/repo/git.go
```diff
@@ -0,0 +1,55 @@
...
+// HTTPClient defines an interface for HTTP get requests.
+//go:generate counterfeiter -o fakes/fake_http_client.go . HTTPClient
+type HTTPClient interface {
+	Get(string) (*http.Response, error)
+}
+
+var httpClient HTTPClient = http.DefaultClient
+
+// GetProfileDefinition returns a definition based on a url and a branch.
+func GetProfileDefinition(repoURL, branch, path string) (profilesv1.ProfileDefinition, error) {
```
Skarlso, line 24: I know this is copied code, but I really really would prefer if that httpClient wouldn't be a package local variable. :D But rather would hang in a struct and this profile would have a struct receiver. Would be easier to mock as well, I believe.

pkg/profile/helm.go
aclevername: just copied this across, but yeah. I think its validated in the profile definition? Idr

pkg/repo/git.go
aclevername: yeah I don't feel strongly either way. I like the pattern of `export_test.go` to "hide" the mocking from the consumer (who wouldn't care what httpClient was passed in). Having it on a struct instead and set the same way is fine, although a struct with 1 func might as well just be a func :smile:

Skarlso: :D, true. :)

Skarlso (approved):
finally read through it. Even though it was mostly copied I wanted to make sure I understand what's being copied. I'm not sure if anything is missing though. :D But code-wise I think it looks good!

PR #64, opened 2021-05-20
Remove pctl get and update list to show catalog source instead of status

aclevername:
### Description

We no longer have the `status.conditions` being set, so thats been removed. It makes sense instead to show the catalog entry a profile was installed from. Also removes the `get` command as `list` shows all the information you need.

```
-> % ./pctl install nginx-catalog/weaveworks-nginx/v0.1.0
generating subscription for profile nginx-catalog/weaveworks-nginx:


-> % k apply -f weaveworks-nginx
gitrepository.source.toolkit.fluxcd.io/pctl-profile-profiles-examples-v0.1.0 created
gitrepository.source.toolkit.fluxcd.io/pctl-profile-profiles-examples-v0.0.1 created
helmrelease.helm.toolkit.fluxcd.io/pctl-profile-bitnami-nginx-nginx-server created
helmrelease.helm.toolkit.fluxcd.io/pctl-profile-weaveworks-nginx-dokuwiki created
helmrepository.source.toolkit.fluxcd.io/pctl-profile-profiles-examples-dokuwiki created
kustomization.kustomize.toolkit.fluxcd.io/pctl-profile-weaveworks-nginx-nginx-deployment created
profilesubscription.weave.works/pctl-profile created

-> % ./pctl list
NAMESPACE       NAME            PROFILE                 VERSION CATALOG
default         pctl-profile    weaveworks-nginx        v0.1.0  nginx-catalog
```

### Checklist


tests/integration/integration_test.go
```diff
@@ -449,7 +396,7 @@ status: {}
 					"--create-pr",
 					"--branch",
 					branch,
-					"nginx-catalog/weaveworks-nginx")
+					"nginx-catalog/weaveworks-nginx/v0.1.0")
```
Skarlso, line 399: We should also have one without version right? Because that also should work. Or do we have one I just don't see it? :D

Skarlso (approved):
LGTM, one question

tests/integration/integration_test.go
aclevername: I think it didn't work actually, which is why I modified it!

Issue #62, opened 2021-05-19
Fix create-pr test


Skarlso:
Right now, the test fails because it's expecting a single file to be there, but that's no longer the case because we are generating multiple artifacts now.

PR #65, opened 2021-05-20
Fixed the create-pr test and introduced out folder to install

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/62

### Checklist

### DIFF
--- a/cmd/pctl/install.go
+++ b/cmd/pctl/install.go
@@ -3,7 +3,6 @@ package main
 import (
 	"errors"
 	"fmt"
-	"path/filepath"
 	"strings"
 
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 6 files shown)


pkg/git/git.go
```diff
@@ -151,10 +149,10 @@ func (g *CLIGit) Push() error {
 func (g *CLIGit) Add() error {
 	fmt.Println("adding unstaged changes")
 	args := []string{
-		"--git-dir", filepath.Join(g.Location, ".git"),
-		"--work-tree", g.Location,
+		"--git-dir", filepath.Join(g.Directory, ".git"),
+		"--work-tree", g.Directory,
 		"add",
-		g.Filename,
+		".",
```
Skarlso, line 155: I don't necessarily like this, but we don't know what files it generated so we kind of need to add anything that is pending after install.

cmd/pctl/install.go
```diff
@@ -59,6 +58,12 @@ func installCmd() *cli.Command {
 				DefaultText: "main",
 				Usage:       "The base branch to open a PR against.",
 			},
+			&cli.StringFlag{
+				Name:        "out",
+				DefaultText: "current",
+				Value:       ".",
+				Usage:       "Optional location to output generated files into.",
+			},
```
aclevername, line 66: its slightly misleading, its "the location to create the profile installation directory in", as opposed to the place where the contents of the profile installation go in

Skarlso: ah, true, that was before I realised it creates a folder. :D

aclevername (approved):
LGTM!

Issue #55, opened 2021-05-18
Add support for pctl install via branch


aclevername:
**What feature/behavior/change do you want?**
Currently `pctl install` only supports deploying via a catalog entry, e.g. `pctl install nginx-catalog/weaveworks-nginx/v0.1.0`. Lets add support for specifying a branch & URL manually. This lets users who are developing profiles use `pctl install` to test the logic.

`pctl install --branch master --url github.com/weaveworks/profiles-examples --path weaveworks-nginx`

**Why do you want this feature?**
To easily develop profiles

aclevername:
need to think about what the new `pctl list` output shows when using branch & path @Skarlso

Skarlso:
```
pctl list
NAME     NAMESPACE     SOURCE
                                           CATALOG                    PROFILE       VERSION
foo          default                nginx-catalog             weaveworks v0.1.0
foo          default                nginx-catalog             weaveworks v0.1.0
foo          default                nginx-catalog             weaveworks v0.1.0

                                 URL                                         BRANCH   PATH
testing  default       github.com/weaveworks/foo main          path/to/profile
```

Skarlso:
Yeah, tablewriter is a pain in the arse. :/ So we settled on this:

```
./pctl list
NAMESPACE       NAME                    SOURCE
default         pctl-profile            nginx-catalog/weaveworks-nginx/v0.1.0
default         pctl-profile-branch     https://github.com/weaveworks/profiles-examples:branch-and-url:branch-nginx
```

PR #69, opened 2021-05-21
Add ability to use branch path and url for developing profiles

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/55

### Checklist


cmd/pctl/install.go
```diff
@@ -67,7 +68,17 @@ func installCmd() *cli.Command {
 			&cli.StringFlag{
 				Name:  "repo",
 				Value: "",
-				Usage: "The repository to open a pr against. Format is: org/repo-name",
+				Usage: "The repository to open a pr against. Format is: org/repo-name.",
```
aclevername, line 71: shall we update the `UsageText` above to demo the two alternative ways of installing a profile?

cmd/pctl/install.go
```diff
@@ -88,24 +99,40 @@ func installCmd() *cli.Command {
...
 	configValues := c.String("config-secret")
 	dir := c.String("out")
-
-	parts := strings.Split(profilePath, "/")
-	if len(parts) < 2 {
-		_ = cli.ShowCommandHelp(c, "install")
-		return errors.New("both catalog name and profile name must be provided")
-	}
-	catalogName, profileName := parts[0], parts[1]
+	path := c.String("path")
```
aclevername, line 135: is this mandatory when url/branch are set?

pkg/catalog/install_test.go
```diff
@@ -163,6 +163,101 @@ status: {}
...
+					Namespace:     "default",
+					ProfileName:   "nginx-1",
+					SubName:       "mysub",
+					URL:           "https://github.com/weaveworks/profiles-examples",
+					Directory:     tempDir,
+					Branch:        "not_domain_compatible",
+					Path:          "path",
+				}
+				err := catalog.Install(cfg)
+				Expect(err).To(MatchError("branch must match RFC 1123 subdomain format"))
```
aclevername, line 258: whats causing this again? It seems odd to block branches that contain `_` as I imagine thats quite common

Skarlso: It's added to the resource name which must be dns compliment

cmd/pctl/install.go
Skarlso: It is otherwise which profile are you installing? Url and branch aren't enough to determine that.

pkg/catalog/install_test.go
aclevername: mhmm, can we strip away the `_`'s and others instead of blocking it? If someones working on a feature PR they shouldn't be prevent from creating the profile because of `_`'s in the branch name

cmd/pctl/install.go
```diff
@@ -67,7 +68,17 @@ func installCmd() *cli.Command {
...
+			&cli.StringFlag{
+				Name:  "url",
+				Value: "",
+				Usage: "Optional value defining the URL of the profile.",
+			},
+			&cli.StringFlag{
+				Name:  "path",
+				Value: "",
+				Usage: "Value defining the path to a profile when url is provided.",
 			},
```
aclevername, line 82: we have an existing `--branch` flag, but thats for `create-pr`, so idk how we want to handle this two different use cases for branchs :cry:

pkg/catalog/install_test.go
Skarlso: Sure, we can convert it.

cmd/pctl/install.go
Skarlso: True. :/ Uh, I'll find something out. :D

pkg/catalog/install_test.go
Skarlso: The problem is that it's not easy. Like, people have all kinds of branches. And stripping out every character is bound to be failure prone. :/ What I'm saying is that it won't be a 100% fool proof coverage.

Skarlso: also, things like `-branch` and `branch--branch` are invalid as well...

aclevername: > The problem is that it's not easy. Like, people have all kinds of branches. And stripping out every character is bound to be failure prone. :/ What I'm saying is that it won't be a 100% fool proof coverage.

yeah... but it feels better than blocking branches with underscores. I'm sure some tool must existing in the k8s community for sanitising names?

alternatively, can we use a different naming convention for the git repository resource that doesn't include the branch name, but is still unique?

cmd/pctl/install.go
Skarlso: done

cmd/pctl/install.go
```diff
@@ -8,15 +8,17 @@ import (
...
 
 func installCmd() *cli.Command {
 	return &cli.Command{
-		Name:      "install",
-		Usage:     "generate a profile subscription for a profile in a catalog",
-		UsageText: "pctl --catalog-url <URL> install --subscription-name pctl-profile --namespace default --branch main --config-secret configmap-name <CATALOG>/<PROFILE>[/<VERSION>]",
+		Name:  "install",
+		Usage: "generate a profile subscription for a profile in a catalog",
+		UsageText: "pctl --catalog-url <URL> install --subscription-name pctl-profile --namespace default --profile-branch main --config-secret configmap-name <CATALOG>/<PROFILE>[/<VERSION>]\n   " +
+			"pctl install --subscription-name pctl-profile --namespace default --profile-branch development --profile-url https://github.com/weaveworks/profiles-examples --profile-path bitnami-nginx",
```
aclevername, line 21: shall we add some more comments? E.g. `to install from profile catalog entry...` , `to install directly from a profile file..`

cmd/pctl/install.go
```diff
@@ -47,27 +49,44 @@ func installCmd() *cli.Command {
...
 				DefaultText: "main",
 				Usage:       "The base branch to open a PR against.",
 			},
+			&cli.StringFlag{
+				Name:        "pr-branch",
+				Value:       "main",
+				DefaultText: "main",
+				Usage:       "The branch to create the PR from.",
+			},
+
```
aclevername, line 69: is it possible to "group" flags with this CLI tool? Would be cool and help readability

Skarlso: will check

cmd/pctl/install.go
Skarlso: sure 👍

cmd/pctl/install.go
```diff
@@ -88,50 +107,64 @@ func installCmd() *cli.Command {
...
+			return errors.New("both catalog name and profile name must be provided")
+		}
+		if len(parts) == 3 {
+			version = parts[2]
+		}
+		catalogName, profileName = parts[0], parts[1]
 	}
 
-	branch := c.String("branch")
+	branch := c.String("profile-branch")
```
aclevername, line 138: shall we error if both are provided?

pkg/profile/artifact_test.go
```diff
@@ -157,10 +157,10 @@ var _ = Describe("Profile", func() {
 			Expect(artifacts).To(HaveLen(4))
 
 			By("generating the nested profile artifact", func() {
-				nestedProfileArifact := artifacts[0]
-				Expect(nestedProfileArifact.Name).To(Equal(filepath.Join(profileName2, chartName1)))
+				nestedProfileArtifact := artifacts[0]
+				Expect(nestedProfileArtifact.Name).To(Equal(filepath.Join(profileName2, chartName1)))
 
-				objects := nestedProfileArifact.Objects
+				objects := nestedProfileArtifact.Objects
```
aclevername, line 163: god dam I can't spell

Skarlso: *damn :D :D :D :D

cmd/pctl/install.go
Skarlso: Nope, nothing that would be convenient and grouping. :/

cmd/pctl/install.go
Skarlso: done

cmd/pctl/install.go
```diff
@@ -88,50 +107,68 @@ func installCmd() *cli.Command {
...
 		Namespace:     namespace,
 		ProfileName:   profileName,
 		SubName:       subName,
 		Directory:     dir,
+		URL:           url,
 		Version:       version,
+		Path:          path,
 	}
 	return catalog.Install(cfg)
 }
```
aclevername, line 164: might do this in a follow up PR, this struct is pretty big. Could maybe be split into:
```
InstallConfig{
  CatalogReference
  ProfileReference
  SubscriptionSpec
}
```

aclevername (approved):

PR #80, opened 2021-05-24
Install latest profile if no version is provided

aclevername:
### Description

We will pass "latest" to the catalog API calls if no version is provided. Sister PR https://github.com/weaveworks/profiles/pull/126

### Checklist

### DIFF
--- a/pkg/catalog/install.go
+++ b/pkg/catalog/install.go
@@ -51,11 +51,11 @@ func Install(cfg InstallConfig) error {
 		},
 		Spec: profilesv1.ProfileSubscriptionSpec{
 			ProfileURL: profile.URL,
-			Version:    filepath.Join(cfg.ProfileName, cfg.Version),
+			Version:    filepath.Join(profile.Name, profile.Version),
 			ProfileCatalogDescription: &profilesv1.ProfileCatalogDescription{
 				Catalog: cfg.CatalogName,
-				Version: cfg.Version,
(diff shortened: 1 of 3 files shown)


pkg/catalog/install.go
```diff
@@ -51,10 +51,10 @@ func Install(cfg InstallConfig) error {
 		},
 		Spec: profilesv1.ProfileSubscriptionSpec{
 			ProfileURL: profile.URL,
-			Version:    filepath.Join(cfg.ProfileName, cfg.Version),
+			Version:    filepath.Join(cfg.ProfileName, profile.Version),
 			ProfileCatalogDescription: &profilesv1.ProfileCatalogDescription{
 				Catalog: cfg.CatalogName,
-				Version: cfg.Version,
+				Version: profile.Version,
 				Profile: cfg.ProfileName,
```
Callisto13, line 58: why not change all these to be `profile.X` i think they all exist there?

doesn't really matter ofc, just wondering why this change is there

Callisto13 (approved):

pkg/catalog/install.go
aclevername: this is because `profile` contains the actual version, e.g. `v0.1.0`. `cfg.Version` is "latest" when not set. I think your right they could all change to `profile.`

Callisto13: ah makes sense

PR #84, opened 2021-05-26
Detect profile update

Skarlso:
### Description

**NOTE**: This PR contains both https://github.com/weaveworks/pctl/pull/69 right now. It will make more sense once 69 is merged.

Closes https://github.com/weaveworks/profiles/issues/101

Depends on Profiles tagging and release for PR https://github.com/weaveworks/profiles/pull/133

### Checklist


pkg/catalog/list.go
```diff
@@ -0,0 +1,59 @@
...
+			if r.Name != p.Profile {
+				continue
+			}
+			rv, err := version.NewVersion(strings.TrimPrefix(r.Version, "v"))
+			if err != nil {
+				return nil, fmt.Errorf("failed to format profile %s with version %s into version: %w", r.Name, r.Version, err)
+			}
+			if rv.GreaterThan(pv) {
+				versions = append(versions, "v"+rv.String())
+			}
```
aclevername, line 51: does it make more sense for this logic to live in the profiles catalog API and to add a new endpoint? Similar to how the "latest" version logic all lives there too

Skarlso: > // Search uses Contains, which will match too many things. We want an exact match.
> // TODO: Maybe delegate this to the profile catalog?

Yep. :D I agree with this question. :D I don't have the answer :) Does it? :D

Skarlso: Oh so sorry. I still can't read apparently. You mean the version compare logic.
I guess? Maybe? I don't know, honestly. It would make the code more simple. It would just say, get GreaterThan and give it a profile name and current version. And then it would return all profiles and versions greater than.

Skarlso: I STILL CAN'T READ. Because you mean 30-51. Frigging github for not highlighting highlighted rows for bigger comments.

Yes, I guess it would make sense to shift this whole logic into the catalog.

aclevername: :coffee: :heavy_check_mark:

Skarlso:
TODOs:

- [ ] Readme update
- [ ] Right now, branch sourced profiles are causing an error on list... they must be excluded from available version check

pkg/catalog/list.go
```diff
@@ -0,0 +1,42 @@
...
+			return nil, fmt.Errorf("failed to search for profile %s for updates: %w", p.Profile, err)
+		}
+		for _, r := range related {
+			versions = append(versions, r.Version)
+		}
+		profileData = append(profileData, ProfileData{
+			Profile:                 p,
+			AvailableVersionUpdates: versions,
+		})
+	}
```
aclevername, line 40: mhmm do we want to error if we can't find available updates, or just put `-`?

Skarlso: It's a perfectly legal scenario. It just won't display anything. :)

Skarlso: Oh yeah, maybe a `-` is better than nothing.

tests/integration/catalog-source.yaml
```diff
@@ -11,5 +11,19 @@ spec:
...
+      maintainer: weaveworks (https://github.com/weaveworks/profiles)
+      prerequisites:
+        - Kubernetes 1.18+
+    - name: multi-version-weaveworks-nginx
+      description: This installs nginx.
+      version: v0.1.1
+      url: https://github.com/weaveworks/profiles-examples
+      maintainer: weaveworks (https://github.com/weaveworks/profiles)
+      prerequisites:
+        - Kubernetes 1.18+
```
aclevername, line 27: shall we also update the `example/profile-source-catalog.yaml` and add a `weaveworks-nginx` `v0.1.1` too? (and create that actual tag, so we don't run into problems :wink: )

aclevername (commented):
LGTM, just missing integration tests and the two comments you pointed out :smile:

tests/integration/catalog-source.yaml
Skarlso: https://github.com/weaveworks/profiles-examples/blob/update-available/weaveworks-nginx/profile.yaml

yeah I've already did something there. :D

Skarlso:
@aclevername What do you mean missing integration test? :D I have an integration test.

tests/integration/catalog-source.yaml
aclevername: need to merge that and cut a tag :smile:

tests/integration/integration_test.go
```diff
@@ -183,9 +183,10 @@ var _ = Describe("PCTL", func() {
 				Expect(err).ToNot(HaveOccurred())
 				return strings.Split(string(session), "\n")
 			}
+
 			Eventually(listCmd).Should(ContainElements(
-				"NAMESPACE	NAME  	SOURCE         ",
-				"default  \tmy-sub\tfoo/bar/v0.1.0\t",
+				"NAMESPACE\tNAME                       \tSOURCE                                   \tAVAILABLE UPDATES ",
+				"default  \tlong-name-to-ensure-padding\tfoo/multi-version-weaveworks-nginx/v0.1.0\tv0.1.1           \t",
```
aclevername, line 189: gah sorry, I focused on the `long-name-to-ensure-padding` and didn't see the update :joy:

tests/integration/catalog-source.yaml
Skarlso: makes sense :)

pkg/catalog/list.go
```diff
@@ -0,0 +1,49 @@
...
+	profileData := make([]ProfileData, 0)
+	for _, p := range profiles {
+		var versions []string
+		// skip for profiles which don't have a catalog entry. i.e.: profiles installed via branch, url, path.
+		if p.Catalog != "-" {
+			related, err := GetAvailableUpdates(catalogClient, p.Catalog, p.Profile, p.Version)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get available updates: %w", err)
+			}
```
aclevername, line 35: do we want to error if we can't get the available updates? We could just put `-` or `n/a` and log an error, but still proceed? Not sure, maybe leave it like this and we can always come back to it? Talking myself out of it haha...

Skarlso: This error is now definitive, like, if the output can't be marshalled. And later on things like, if the catalog doesn't respond. I think we want to error in those cases? The rest are ignored and `-` is used.

If we output something, we might break the formatter so I don't want to do that. That could mess with the table / json output.

tests/integration/integration_test.go
```diff
@@ -183,11 +207,28 @@ var _ = Describe("PCTL", func() {
...
+		//			Expect(err).ToNot(HaveOccurred())
+		//			return strings.Split(string(session), "\n")
+		//		}
+		//
+		//		Eventually(listCmd).Should(ContainElements(
+		//			"NAMESPACE\tNAME                       \tSOURCE                                   \tAVAILABLE UPDATES ",
+		//			"default  \tlong-name-to-ensure-padding\tfoo/multi-version-weaveworks-nginx/v0.1.0\tv0.1.1           \t",
+		//		))
+		//	})
+		//})
```
aclevername, line 231: :eyes:

Skarlso: Ah! Thanks . :D I wanted to finish this test. But I ran out of time yesterday. :) I also want to add a case in which there is no new version. :)

aclevername (approved):

Issue #75, opened 2021-05-24
Generated artifact files should not be 0755 


Callisto13:
Run `pctl install <foo>` then do `ls -l`. The generated files are all `0755`. We should be generating `0644`s.

aclevername:
@Callisto13 correcting your titles for a change :grin:

Callisto13:
The british spelling has an `e` 😏  so blame chrome spellcheck

Callisto13:
Wont be at standup, but this is what I will be doing when I get back later.

Given in the code we are creating with `0777`, if i were to hazard a guess I would say that there is a `umask` of `0022` at play somewhere.

PR #85, opened 2021-05-26
Generate artifact files with 0644 perms

Callisto13:
### Description

We previously created with `0777` which, depending on the caller's local
`umask`, would result in files created with `0755` or `0775` (assuming a
typical `umask` setting of `0022` or `0002`).

This commit chooses `0644` as a suitably permissive mode for files. `0664`
would also have been okay, but again this is more likely to be
affected by individual umasks so is less predictable.

Closes https://github.com/weaveworks/pctl/issues/75

### Checklist


Skarlso (approved):
LGTM

PR #89, opened 2021-05-28
Rename files in pkg/repo

Callisto13:
### Description

These files were called `git*` while the package is called `repo`, and since we already have another `git` package (and test suite) this is less confusing.


Skarlso (approved):
LGTM

Issue #104, opened 2021-06-02
pctl install copies down remote artifacts


aclevername:
**What feature/behavior/change do you want?**
Follow up from https://github.com/weaveworks/pctl/issues/103 we added support for installing a profile from a private repository. Now we have a problem that we reference artifacts in that remote repository, which is private. One solution is to ask the user to provide a secret that can be used to clone it, or we can clone those artifacts locally into the profile installation. We are going for the later.

## Example
Lets say `github.com/private/dokuwiki` exists and contains the `dokuwiki` profile which consists of a locally defined helm chart:

```
apiVersion: profiles.fluxcd.io/v1alpha1
kind: Profile
metadata:
  name: dokuwiki
spec:
  description: Profile for deploying local dokuwiki
  version: v0.0.1
  artifacts:
    - name: dokuwiki
      path: dokuwiki/chart
      kind: HelmChart
```

A user installs it via `pctl install my-catalog/dokuwiki`
```
├── dokuwiki
│   ├── GitRepository.yaml # pointing to github.com/private/dokuwiki to consume the local chart
│   └── HelmRelease.yaml
```
That GitRepository resource points to `github.com/private/dokuwiki` and is referenced in the `HelmRelease` resource

```yaml
apiVersion: source.toolkit.fluxcd.io/v1beta1
kind: GitRepository
metadata:
  name: bitnami-profiles-examples-main
  namespace: default
spec:
  interval: 0s
  ref:
    branch: main
  url: https://github.com/private/dokuwiki
```

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: bitnami-bitnami-nginx-nginx-server
  namespace: default
spec:
  chart:
    spec:
      chart: dokuwiki/chart
      sourceRef:
        kind: GitRepository
        name: bitnami-profiles-examples-main
        namespace: default
```

We now instead want to copy down the `chart` definition, and have re-use the existing `GitRepository` resource (since the profile installation repo is already watched by some GitRepository resource)

## New installation
`pctl install dokuwiki --git-repository=<namespace>/<name>`
```
dokuwiki
├── chart
│   ├── Chart.yaml
│   └── templates
│       └── template-1.yaml
└── HelmRelease.yaml
```

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: bitnami-bitnami-nginx-nginx-server
  namespace: default
spec:
[...]

Issue #115, opened 2021-06-08
Refactor pctl profiles global functions into providers and builders


Skarlso:
Right now, there are these global functions all over the place:

```go
//MakeArtifacts returns artifacts for a subscription
type MakeArtifacts func(sub profilesv1.ProfileSubscription, gitClient git.Git)([]profile.Artifact, error)
```

These functions make progress difficult as every change we would like to make to generating artifacts needs to change the function definition. Which doesn't scale really well. 

We need to create some interfaces to abstract this behaviour. And while we are there, we also need some builders abstraction on the Kind switch so we can compartmentalise those changes and make them easy to extend once we inevitably start expending them.

This should be done sooner rather than later, because with every feature it will get more difficult to do.

Callisto13:
@Skarlso can this be closed?

Skarlso:
Oh yes. Thanks!

PR #119, opened 2021-06-09
Clone repository local artifacts so they are available for the flux operator

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/104

### Checklist


tests/integration/prepare_test_cluster.go
```diff
@@ -51,7 +51,7 @@ func PrepareTestCluster(binaryPath string) error {
 			fmt.Printf("failed to remove temporary folder at location: %s. Please clean manually.", tmp)
 		}
 	}()
-	cmd := exec.Command(binaryPath, "prepare", "--dry-run", "--out", tmp, "--keep")
+	cmd := exec.Command(binaryPath, "prepare", "--version", "v0.0.2", "--dry-run", "--out", tmp, "--keep")
```
Skarlso, line 54: @Skarlso Remove this pin

README.md
```diff
@@ -133,12 +133,35 @@ pctl install --subscription-name pctl-profile \
...
+             --namespace [default] \
+             --profile-branch [main] \
+             --profile-url git@github.com:org/private-profile-repo \
+             --profile-path <profile-name> \
+             --git-repository <namespace>/<name> \
+             --out <location of my flux repository>
+```
+
+This will result in all the private resources which aren't available from the outside, downloaded locally and added
+into the flux repository.
```
aclevername, line 163: doesn't matter if its public right? still gets cloned

pkg/profile/git_repo.go
```diff
@@ -1,25 +1,25 @@
...
 	ref := &sourcev1.GitRepositoryRef{
-		Branch: p.subscription.Spec.Branch,
+		Branch: fmt.Sprintf("%s:%s:%s", p.definition.Name, path, p.subscription.Spec.Branch),
 	}
 	if p.subscription.Spec.Tag != "" {
 		ref = &sourcev1.GitRepositoryRef{
-			Tag: p.subscription.Spec.Tag,
+			Tag: fmt.Sprintf("%s:%s:%s", p.definition.Name, path, p.subscription.Spec.Tag),
 		}
 	}
```
aclevername, line 18: slightly confused about this, are we just using this resource as a convenient way to pass information around?

pkg/catalog/install.go
```diff
@@ -88,12 +121,53 @@ func Install(cfg InstallConfig, gitClient git.Git) error {
 	return generateOutput(filepath.Join(profileRootdir, "profile.yaml"), &subscription)
 }
 
+// cloneGitRepositoryObjects takes a GitRepository objects and clones what it points to, then discards the object.
+func cloneGitRepositoryObjects(cfg InstallConfig, g *sourcev1.GitRepository, artifactDir string) error {
```
aclevername, line 125: does this mean we do a `git clone` per artifact that we need to copy? also the naming is confusing, wb `getGitRepositoryArtifact` or something, the fact that it clones doesn't matter :smile:

pkg/catalog/install.go
```diff
@@ -88,12 +121,53 @@ func Install(cfg InstallConfig, gitClient git.Git) error {
...
+	if strings.Contains(path, string(os.PathSeparator)) {
+		path = filepath.Dir(path)
+	}
+	fullPath := filepath.Join(tmp, profilePath, path)
+	if err := os.Rename(fullPath, filepath.Join(artifactDir, path)); err != nil {
+		return fmt.Errorf("failed to move folder: %w", err)
+	}
+	return nil
+}
+
```
aclevername, line 167: we missing some repo cleanup?

README.md
```diff
@@ -133,12 +133,35 @@ pctl install --subscription-name pctl-profile \
...
 ```
 
 It's the user's responsibility to make sure that the local `git` setup has access to the url provided with `profile-url`.
 It can be any form of url as long as `git clone` understands it.
 
+#### Cloning private repository resources
+
+pctl clones all repository local resources and puts them into the flux repository. This is done so that the user doesn't
+have to include access credentials for all private repositories including nested profiles. However, for repository local
+resources to work, the user has to provide the location of the GitRepository object that flux manages.
```
aclevername, line 146: I feel like this line doesn't make it super clear that "provide us the gitrepository resource that points to the place where your installing your profile", but I have no idea how to word this nicely

pkg/profile/artifact.go
```diff
@@ -88,22 +91,30 @@ func (p *Profile) makeArtifacts(profileRepos []string, gitClient git.Git) ([]Art
...
 			artifacts = append(artifacts, a)
 		case profilesv1.KustomizeKind:
-			a.Objects = append(a.Objects, p.makeKustomization(artifact, profileRepoPath))
-			a.Objects = append(a.Objects, p.makeGitRepository())
+			if p.gitRepositoryNamespace == "" && p.gitRepositoryName == "" {
+				return nil, fmt.Errorf("in case of local resources, the flux gitrepository object's details must be provided")
+			}
+			path := filepath.Join(p.rootDir, "artifacts", artifact.Name, artifact.Path)
+			a.Objects = append(a.Objects, p.makeKustomization(artifact, path))
+			a.Objects = append(a.Objects, p.makeGitRepository(artifact.Path))
```
aclevername, line 117: could we split all of the contents of each `case` into a `func GenerateXArtifact` or somthing? This was already too big :smile:

tests/integration/catalog-source.yaml
```diff
@@ -30,8 +30,8 @@ spec:
         - Kubernetes 1.18+
     - name: nginx
       description: This installs nginx.
-      version: v1.0.0
-      tag: v1.0.0
+      version: v2.0.0
+      tag: v2.0.0
```
aclevername, line 34: nice

README.md
Skarlso: I'll figure something out.

pkg/profile/artifact.go
Skarlso: I didn't want to do that because https://github.com/weaveworks/pctl/issues/115 proposes to use builders in any case.

tests/integration/catalog-source.yaml
Skarlso: :)

pkg/catalog/install.go
Skarlso: Nope, it's defer removed at the begin.

aclevername: I am blind.

aclevername (approved):

PR #130, opened 2021-06-11
fix pended unit test

aclevername:
### Description

Moved the validate logic to here, since its where we consume and test that logic. LMK if folks would rather move it, in which case the testing should move too

Sister PR in profiles https://github.com/weaveworks/profiles/pull/158

### Checklist


Skarlso (approved):
LGTM

PR #136, opened 2021-06-14
Extracting the MakeArtifacts function into a provider

Skarlso:
### Description

The first step in doing
https://github.com/weaveworks/pctl/issues/115

### This PR

This PR introduces the ArtifactsMaker interface. From here on, we'll start slowly extracting the artifacts making into builders and eventually determine the responsibilities of profiles and where things should sit.

Further:
Introduce individual builders, and determine where `makeArtifacts` function should sit. As the name suggests, it should be the responsibility of the artifacts maker with a `profiles` provider which takes care of the details.

Also extract the GitClient from `Install`. It should be on the MakeArtifacts only, `Install` shouldn't care about the details of what's happening there and how the artifacts are created and output. It should only call its providers to handle generated data.

### Checklist


pkg/catalog/install_test.go
```diff
@@ -67,38 +67,37 @@ var _ = Describe("Install", func() {
...
+			},
+		}, nil)
 	})
 
-	JustBeforeEach(func() {
-		catalog.SetMakeArtifacts(fakeMakeArtifacts)
-	})
+	//JustBeforeEach(func() {
+	//	catalog.SetMakeArtifacts(fakeMakeArtifacts)
+	//})
```
aclevername, line 101: ?

Skarlso: ops, left that in there, I want to delete it, thanks

aclevername (approved):

PR #140, opened 2021-06-14
don't print table columns when no results found for list

aclevername:
Before:
```
pctl list
no profiles found
NAMESPACE       NAME    SOURCE  AVAILABLE UPDATES
```

After:

```
./pctl list
no profiles installed
```


Skarlso (commented):
LGTM

Skarlso (approved):

PR #142, opened 2021-06-14
Shifting the artifacts generation to the artifact maker

Skarlso:
### Description

https://github.com/weaveworks/pctl/issues/115

### In This PR

Shift the generation of the output into the artifact maker, lessening the dependencies of `Install`.

### Checklist

### DIFF
--- a/pkg/catalog/install.go
+++ b/pkg/catalog/install.go
@@ -2,19 +2,11 @@ package catalog
 
 import (
 	"fmt"
-	"io/ioutil"
-	"os"
-	"path/filepath"
 	"strings"
 
 	helmv2 "github.com/fluxcd/helm-controller/api/v2beta1"
-	"github.com/google/uuid"
 	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	kjson "k8s.io/apimachinery/pkg/runtime/serializer/json"
-
-	"github.com/otiai10/copy"
 
 	"github.com/weaveworks/pctl/pkg/git"
 	"github.com/weaveworks/pctl/pkg/profile"
@@ -23,7 +15,6 @@ import (
 // Clients contains a set of clients which are used by install.
 type Clients struct {
 	CatalogClient  CatalogClient
-	GitClient      git.Git // this will be refactor out of here, so it's not Install's responsibility
 	ArtifactsMaker profile.ArtifactsMaker
 }
 
@@ -38,14 +29,12 @@ type ProfileConfig struct {
 	ProfileBranch string
 	URL           string
 	Path          string
-	GitRepository string
 }
 
 // InstallConfig defines parameters for the installation call.
 type InstallConfig struct {
 	Clients
 	ProfileConfig
-	Directory string
 }
 
 // Install using the catalog at catalogURL and a profile matching the provided profileName generates a profile subscription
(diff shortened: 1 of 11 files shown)


pkg/catalog/install.go
```diff
@@ -75,64 +64,12 @@ func Install(cfg InstallConfig) error {
...
-	for _, path := range a.PathsToCopy {
-		// nginx/chart/...
-		if strings.Contains(path, string(os.PathSeparator)) {
-			path = filepath.Dir(path)
-		}
-		fullPath := filepath.Join(tmp, profilePath, path)
-		if err := copy.Copy(fullPath, filepath.Join(artifactDir, path)); err != nil {
-			return fmt.Errorf("failed to move folder: %w", err)
-		}
+	if err := cfg.ArtifactsMaker.GenerateArtifactsOutput(artifacts, subscription); err != nil {
```
aclevername, line 71:
```suggestion
	if err := cfg.ArtifactsMaker.Generate(artifacts, subscription); err != nil {
```

or something like that, to reduce repetetion

pkg/profile/artifact.go
```diff
@@ -49,6 +56,84 @@ func NewProfilesArtifactsMaker(cfg MakerConfig) *ProfilesArtifactsMaker {
 	}
 }
 
+// GenerateArtifactsOutput takes a slice of generated artifacts and creates files, folders and cloned artifacts for them.
+func (pa *ProfilesArtifactsMaker) GenerateArtifactsOutput(artifacts []Artifact, profilesInstallations profilesv1.ProfileInstallation) error {
```
aclevername, line 60: Do we need `MakeArtifacts` and `GenerateArtifactsOutput` as two separate functions? They only ever get used in combination right? I know its easier for testing but it doesn't really serve much purpose. WDYT?

pkg/profile/artifact_test.go
```diff
@@ -678,4 +681,237 @@ var _ = Describe("Profile", func() {
...
+				ProfileName:      "generate-test",
+				GitClient:        fakeGitClient,
+				RootDir:          tempDir,
+				GitRepoNamespace: gitRepoNamespace,
+				GitRepoName:      gitRepoName,
+			})
+			artifacts, err := maker.MakeArtifacts(pSub)
+			Expect(err).NotTo(HaveOccurred())
+			err = maker.GenerateArtifactsOutput(artifacts, pSub)
+			Expect(err).NotTo(HaveOccurred())
```
aclevername, line 854: if we do keep these funcs separate its better to not test one functions in another functions suite, so instead we could fake out the `Artifacts` struct. Might also make the test smaller/cleaner :smile:

pkg/profile/artifact.go
Skarlso: Thought about that too, yes. I don't know. 😁 I kept it separate in the end because I thought about things like, we might want to keep the way we generate artifacts but change how we output them?

pkg/profile/artifact_test.go
Skarlso: Yeah, agreed

pkg/profile/artifact_test.go
```diff
@@ -678,4 +681,237 @@ var _ = Describe("Profile", func() {
...
+			})
+			artifacts, err := maker.MakeArtifacts(pSub)
+			Expect(err).NotTo(HaveOccurred())
+			fakeGitClient.SparseCloneStub = func(url string, branch string, dir string, p string) error {
+				fullPath := filepath.Join(dir, "weaveworks-nginx", "nginx", "deployment")
+				err := os.MkdirAll(fullPath, 0755)
+				Expect(err).NotTo(HaveOccurred())
+				return ioutil.WriteFile(filepath.Join(fullPath, "deployment.yaml"), []byte("validYaml:"), 0644)
+			}
+			err = maker.GenerateOutput(artifacts, pSub)
```
aclevername, line 741: I think this naming is slightly messy. We have an `ArtifactMaker` (whats make in this context? the in memory structs, the files?), which has a `MakeArifacts` (can probably just be `Make` or really in an ideal world just `()` since its implied from struct name). And the `GenerateOutput` which feels misplaced.

I think part of the smell is that its two funcs that should be one, which makes naming the functions and the struct hard. So how about either:
- Moving the file writing functionality outside of this interface, which is probably not ideal since it was moved here in this PR aha :wink: but perhaps theres a slightly different abstraction that games the generating"and writing to files separate
- Merge them to one func, then we have `ArtifactMaker.Make()` which keeps it a bit cleaner. Then we can extract out some underlying parts of the ArtifactMaker into smaller units? wdyt?

Callisto13:
I feel like there was a missed opportunity to call the `ArtifactMaker` the `Artifactory` :trollface:

aclevername:
> I feel like there was a missed opportunity to call the `ArtifactMaker` the `Artifactory` :trollface:

:man_facepalming:

pkg/profile/artifact_test.go
Callisto13: Struct: `Artifactory` (yes it is an epic idea)
Funcs: `Make` and `Write` (isn't that what it is doing? I don't get where the "generate" comes from)

Skarlso: Oh, sorry Claudia, I didn't see your comment. :( I put everything into a single `Make` in the end.

Callisto13: No worries 😄  can I ask why you chose to refactor at this level, around the general "make all the things call"? I assumed you would be having the interface at the `makeArtifacts` level and then have each artifact fulfil the interface and get rid of that giant `if else` which is growing by the second.

And I realise I should have asked this in the first PR 🤦‍♀️

Skarlso: `makeArtifact` will disappear and profile will be a provider for the Maker. And then profiles will have builders which build the individual artifacts. That's the main idea. :)

Callisto13: Perfect!

tests/integration/integration_test.go
```diff
@@ -699,12 +699,12 @@ status: {}
 
 			By("creating the artifacts")
 			Expect(files).To(ContainElements(
-				"profile-installation.yaml",
-				"artifacts/bitnami-nginx/HelmRelease.yaml",
-				"artifacts/bitnami-nginx/HelmRepository.yaml",
+				"nginx/profile-installation.yaml",
```
Skarlso, line 702: this was missing, because this is using a an added subfolder

pkg/catalog/install_test.go
```diff
@@ -103,161 +73,47 @@ var _ = Describe("Install", func() {
...
-    name: ""
-status: {}
-`))
+				Expect(err).To(MatchError("failed to make artifacts: foo"))
 			})
 		})
 
 		When("a branch is provided which isn't domain compatible", func() {
 			It("will not care because the name is sanitised", func() {
+				fakeMakeArtifacts.MakeReturns(nil)
```
Skarlso, line 113: I don't think this is needed here

pkg/profile/artifact_test.go
```diff
@@ -130,276 +123,143 @@ var _ = Describe("Profile", func() {
...
+  interval: 5m0s
+  path: %s/artifacts/nginx-deployment/nginx/deployment
+  prune: true
+  sourceRef:
+    kind: GitRepository
+    name: git-repo-name
+    namespace: git-repo-namespace
+  targetNamespace: default
+status: {}
+`, rootDir)))
```
aclevername, line 235: we need to test all the other artifacts are correct too

Skarlso: What do you mean? Like the others a couple lines below this one?

Skarlso: And the profile-installation. What... the... heck. I thought I'm testing that.

Skarlso: Added all the things!

aclevername: :smile:

aclevername (approved):
LGTM. I think a future refactor (which I might do :wink: ) is to refactor the `artifact_test` to instead of being 1 big profile with all 4 artifacts types, 4 different tests with its own profile with 1 artifact in it each. Will make the test smaller and easier to read

Issue #58, opened 2021-05-19
As a Profile Author I want to be able to provide a pre-configured set of helm values for a helm artifact


aclevername:
As a Profile Author I want to be able to provide a pre-configured set of helm values for a helm artifact. 

Lets say I'm a big fan of the [bitnami-nginx chart](https://github.com/bitnami/charts/tree/master/bitnami/nginx) and I am creating a profile that has it as an artifact. This chart has hundreds of configuration options, but as the profile author I have some opinions about some of these values and want to configure them myself, removing the need for the user to do so themselves. Lets say by default my company wants all components to omit promethues metrics. This is currently a configuration option in the helm chart values file:
```yaml
## Prometheus Exporter / Metrics
##
metrics:  
  enabled: false
```

Rather than having everyone that consumes this profile configure this, I want to be able to specify in my profile definition that `metrics.enabled` should be `true`

## Example
Given a profile:

``` bash
$ tree weaveworks-nginx/
├── profile.yaml
```

`profile.yaml` contains:
```yaml
apiVersion: profiles.fluxcd.io/v1alpha1
kind: Profile
metadata:
  name: weaveworks-nginx
spec:
  description: Profile for deploying nginx
  version: v0.1.0
  artifacts:
    - name: nginx
      kind: HelmChart
      chart:
        url: https://charts.bitnami.com/bitnami
        name: nginx
        version: "11.1.6"
```

I should be able to define some pre-existing helm values that should be applied correctly during `pctl install`. 

## Option 1 - inline values
I can provide a `values` section in my artifact, that get consumed during `pctl install`. For example the new `profile.yaml` contains:

```yaml
apiVersion: profiles.fluxcd.io/v1alpha1
kind: Profile
metadata:
  name: weaveworks-nginx
spec:
  description: Profile for deploying nginx
  version: v0.1.0
  artifacts:
    - name: nginx
      kind: HelmChart
      chart:
        url: https://charts.bitnami.com/bitnami
        name: nginx
        version: "11.1.6"
      values:
         metrics:
             enabled: true
```

## Option 2 - values file artifact
I can provide a `values.yaml` artifact that gets consumed during `pctl install`. For example the new `profile.yaml` contains:

```yaml
apiVersion: profiles.fluxcd.io/v1alpha1
kind: Profile
metadata:
  name: weaveworks-nginx
spec:
  description: Profile for deploying nginx
  version: v0.1.0
  artifacts:
    - name: nginx
      kind: HelmChart
[...]

PR #143, opened 2021-06-15
Create ConfigMap Artifact for chart default values

Callisto13:
### Description

Still WIP

1. Removed `Values` wiring; this is not exposed via `pctl` anyway, so we can ditch it until we need it.
2. Create ConfigMap Artifact for chart default values. If the profile author has set default values on a Chart Artifact, we create a ConfigMap and append it to the Helm Objects array. The reference to the ConfigMap is then placed in the `ValuesFrom` array in first position on the HelmRelease Artifact. Any user configuration is then appended from index 1 onwards. This will ensure any user values override the author's default ones.

Coming soon:
- Complementary PRs in other repos
  - Profiles: https://github.com/weaveworks/profiles/pull/163
  - Examples: https://github.com/weaveworks/profiles-examples/pull/7
  - Test example: https://github.com/weaveworks/nginx-profile/pull/12
- Docs
- Fixing the whole user-provided ConfigMap thing because it was quite buggy the last time I tried to use it

Closes https://github.com/weaveworks/pctl/issues/58

### Checklist


pkg/profile/helm.go
```diff
@@ -81,3 +95,26 @@ func (p *Profile) makeHelmChartSpec(chart string, version string) helmv2.HelmCha
 		Version: version,
 	}
 }
+
+func (p *Profile) makeDefaultValuesCfgMap(name, data string) *corev1.ConfigMap {
+	return &corev1.ConfigMap{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "ConfigMap",
+			APIVersion: "v1",
```
Callisto13, line 103: anyone know if these values exist somewhere I can use?

Skarlso: I'm not sure I understand the question, sorry.... What do you mean exists somewhere..?

pkg/profile/helm.go
```diff
@@ -32,13 +36,24 @@ func (p *Profile) makeHelmRepoName(name string) string {
...
 	var helmChartSpec helmv2.HelmChartTemplateSpec
 	if artifact.Chart.Path != "" {
 		helmChartSpec = p.makeGitChartSpec(path.Join(repoPath, artifact.Chart.Path))
 	} else if artifact.Chart != nil {
 		helmChartSpec = p.makeHelmChartSpec(artifact.Chart.Name, artifact.Chart.Version)
 	}
+
+	var values []helmv2.ValuesReference
+	if defaultValuesRef != "" {
+		values = []helmv2.ValuesReference{{
```
Skarlso, line 49: How come this isn't an `append`?

Callisto13: it could be. it is also the first time the values array is assigned a value, so there is nothing in it 🤷‍♀️

pkg/profile/helm.go
Callisto13: so like for other types we use constants, eg:
```go
			Kind:       sourcev1.HelmRepositoryKind,
			APIVersion: sourcev1.GroupVersion.String(),
```

Skarlso: Oh, I see! It's `metav1` :)

Skarlso: `	v1 "k8s.io/api/core/v1"`

Callisto13: hmmm no it has to be a string. i don;t think it exists?

Skarlso: Oh yes! You're right. Yeah, the core ones don't have Kind defined as a string.

pkg/profile/helm.go
Skarlso: yeah, fare enough ¯\_(ツ)_/¯

pkg/profile/artifact.go
```diff
@@ -206,7 +207,16 @@ func (p *Profile) makeArtifacts(profileRepos []string, gitClient git.Git) ([]Art
...
-			helmRelease := p.makeHelmRelease(artifact, profileRepoPath)
+			var (
+				cfgMap  *corev1.ConfigMap
+				mapName string
+			)
+			if artifact.Chart.DefaultValues != nil {
+				cfgMap = p.makeDefaultValuesCfgMap(artifact.Name, *artifact.Chart.DefaultValues)
+				mapName = cfgMap.ObjectMeta.Name
+				a.Objects = append(a.Objects, cfgMap)
+			}
```
aclevername, line 218: could this maybe move to the `makeHelmRelease` func?

Callisto13: hmmmm i did have it there for a minute. why did i move it? 🤔

aclevername: to troll me?

Callisto13: okay i can't remember.

So i can have `makeHelmRelease` return a `[]runtime.Object` which will get added to the objects list. But it means I will have to do a `helmRelease := objects[0].(*helmv2.HelmRelease)` type thing so that we can do `helmRelease.Spec.Chart.Spec.Chart = filepath.Join....` a couple lines after.

Or I can have `makeHelmRelease` return 2 definite types, and do if nil checks around adding the cfg map to the objects list. Second one is probably nicer and less brittle i think

Callisto13: > to troll me?

I am not saying that is why, I am also not saying it is not why

Skarlso: Oh Gods, I'm gonna have a great time merging this. :D

aclevername: Revenge is a beautiful thing :trollface:

Callisto13: oh but now what do i name this func 🤔

aclevername: > oh but now what do i name this func

`makeHelmReleaseObjects`?

Callisto13: sure if you want to be boring i guess

tests/integration/integration_test.go
```diff
@@ -699,9 +699,10 @@ status: {}
 
 			By("creating the artifacts")
 			Expect(files).To(ContainElements(
-				"nginx/profile-installation.yaml",
-				"nginx/artifacts/bitnami-nginx/HelmRelease.yaml",
-				"nginx/artifacts/bitnami-nginx/HelmRepository.yaml",
+				"artifacts/bitnami-nginx/ConfigMap.yaml",
+				"profile-installation.yaml",
+				"artifacts/bitnami-nginx/HelmRelease.yaml",
+				"artifacts/bitnami-nginx/HelmRepository.yaml",
```
aclevername, line 705: is this missing a check to see if the configmaps correct?

aclevername (approved):
LGTM, 1 comment

tests/integration/integration_test.go
Callisto13: 🤔  i can probably grab the deployed pod and check that the applied spec is correct? does feel a bit like we are testing a well-known feature of theirs tho

Callisto13: i will think about it, and merge what we have now so you are unblocked

aclevername: > thinking i can probably grab the deployed pod and check that the applied spec is correct? does feel a bit like we are testing a well-known feature of theirs tho

True... but not testing it also sort of feels wrong :shrug:

PR #144, opened 2021-06-16
Refactor using builders

Skarlso:
### Checklist


pkg/profile/artifact.go
```diff
@@ -46,28 +47,35 @@ type MakerConfig struct {
...
+
+	Builders     map[int]Builder
+	nestedName   string
+	profileRepos []string
 }
 
 // NewProfilesArtifactsMaker creates a new profiles artifacts maker.
 func NewProfilesArtifactsMaker(cfg MakerConfig) *ProfilesArtifactsMaker {
+	builders := map[int]Builder{
+		KUSTOMIZE: &KustomizeBuilder{
```
Skarlso, line 59: Yeah this could be some kind of shared config, like builder config. I haven't gotten that far yet. :)

pkg/profile/artifact.go
```diff
@@ -94,6 +102,73 @@ func (pa *ProfilesArtifactsMaker) Make(installation profilesv1.ProfileInstallati
...
+			artifact.Name = filepath.Join(pa.nestedName, artifact.Name)
+		}
+
+		if err := validateArtifact(artifact); err != nil {
+			return nil, fmt.Errorf("validation failed for artifact %s: %w", artifact.Name, err)
+		}
+
+		t := -1
+		if artifact.Profile != nil {
+			profileRepoName := profileRepo(installation)
```
Skarlso, line 128: I realised that in fact, we only have two builders. Which is Kustomize and Chart. In any other case, it's just a recursive call. So I don't have a builder for "Profiles".

This makes this part a bit more verbose though, so I'm thinking about what to do about that. :)

pkg/profile/sanitise_branch.go
```diff
@@ -1,14 +0,0 @@
-package profile
```
Skarlso, line 1: Since we don't create GitRepository any longer, these are not used.

pkg/profile/chart_builder.go
```diff
@@ -0,0 +1,116 @@
...
+				Spec: helmChartSpec,
+			},
+			Values:     installation.Spec.Values,
+			ValuesFrom: installation.Spec.ValuesFrom,
+		},
+	}
+	return helmRelease
+}
+
+func (c *ChartBuilder) makeHelmRepository(url string, name string, installation profilesv1.ProfileInstallation) *sourcev1.HelmRepository {
```
Skarlso, line 73: In some cases, installation is not needed, only the namespace. I'm fixing passing around the installation and definition too much.

pkg/profile/artifact.go
```diff
@@ -94,6 +103,61 @@ func (pa *ProfilesArtifactsMaker) Make(installation profilesv1.ProfileInstallati
...
+		} else if artifact.Chart != nil {
+			t = CHART
+		}
+		if t == -1 {
+			return nil, errors.New("no artifact set")
+		}
+		arts, err := pa.Builders[t].Build(artifact, installation, definition)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build artifact: %w", err)
+		}
```
aclevername, line 155: I find `t` super confusing, wdyt about this refactor:
```suggestion
		} else if artifact.Kustomize != nil {
			builder = pa.Builders[KUSTOMIZE]
		} else if artifact.Chart != nil {
			builder = pa.Builders[CHART]
		} else {
			return nil, errors.New("no artifact set")
		}
		arts, err := builder.Build(artifact, installation, definition)
		if err != nil {
			return nil, fmt.Errorf("failed to build artifact: %w", err)
		}
```

pkg/profile/artifact.go
```diff
@@ -94,6 +103,61 @@ func (pa *ProfilesArtifactsMaker) Make(installation profilesv1.ProfileInstallati
...
+			}
+			pa.profileRepos = append(pa.profileRepos, profileRepoName)
+			nestedArtifacts, err := pa.handleNestedProfile(artifact, installation)
+			if err != nil {
+				return nil, err
+			}
+			artifacts = append(artifacts, nestedArtifacts...)
+			pa.nestedName = ""
+			pa.profileRepos = nil
+			continue
```
aclevername, line 143: could this be its own `builder`? or is that too messy :smile:

pkg/profile/artifact_test.go
```diff
@@ -611,120 +622,11 @@ status: {}
 						GitRepoName:      gitRepoName,
 					})
 					err := maker.Make(pSub)
-					Expect(err).To(MatchError(ContainSubstring("failed to generate resources for nested profile \"bitnami-nginx\": failed to generate resources for nested profile \"bitnami-nginx/recursive\": failed to generate resources for nested profile \"bitnami-nginx/recursive/recursive\": recursive artifact detected: profile https://github.com/org/repo-name-nested on branch")))
+					Expect(err).To(MatchError(ContainSubstring("recursive artifact detected: profile example.com/nested on branch main contains an artifact that points recursively back at itself")))
 				})
 			})
 		})
 	})
-	When("there is a single profile repository", func() {
```
aclevername, line 619: this file is still 600 lines?!!?! Oh cmon! :sob:

Skarlso: Yeah I know. :/ it's mostly setup now... and errors. I don't know what to do with it! :( I could put every test into separate files. :D I don't know. :/

pkg/profile/artifact_single_repo_test.go
```diff
@@ -0,0 +1,143 @@
+package profile_test
```
aclevername, line 1: how come we moved this test to a separate file?

pkg/profile/artifact.go
Skarlso: Waay too messy because it needs to know about the other builders and/or the maker which is a circular dependency.
And essentially, it's not a builder... It's just a recursive call to make :)

pkg/profile/artifact_single_repo_test.go
Skarlso: To make artifact_test shorter. :D

pkg/profile/chart_builder.go
```diff
@@ -0,0 +1,116 @@
...
+		branch := installation.Spec.Source.Branch
+		if installation.Spec.Source.Tag != "" {
+			branch = installation.Spec.Source.Tag
+		}
+		a.RepoURL = installation.Spec.Source.URL
+		a.SparseFolder = definition.Name
+		a.Branch = branch
+		a.PathsToCopy = append(a.PathsToCopy, artifact.Chart.Path)
+	}
+	if artifact.Chart.URL != "" {
```
aclevername, line 39: should this be an else?

pkg/profile/chart_builder.go
```diff
@@ -0,0 +1,116 @@
...
+	if artifact.Chart.Path != "" {
+		if c.GitRepositoryNamespace == "" && c.GitRepositoryName == "" {
+			return nil, fmt.Errorf("in case of local resources, the flux gitrepository object's details must be provided")
+		}
+		helmRelease.Spec.Chart.Spec.Chart = filepath.Join(c.RootDir, "artifacts", artifact.Name, artifact.Chart.Path)
+		branch := installation.Spec.Source.Branch
+		if installation.Spec.Source.Tag != "" {
+			branch = installation.Spec.Source.Tag
+		}
+		a.RepoURL = installation.Spec.Source.URL
```
aclevername, line 34: big nit: I think we are missing a unit test for this `if` statement

Skarlso: ahh, I only added it to kustomize builder.
```go
		When("branch is defined instead of tag", func() {
			It("will use the branch definition", func() {
```
:D

aclevername (commented):
Love it man! excellent stuff. Left a few comments but looks like a great refactor!

pkg/profile/artifact_test.go
aclevername: isn't the majority of this file duplicating testing the other builder tests? I think a lot of it could get deleted, and replaced with `fake` builders to reduce the cost and make it easier to make changes to the builders without having to change these test. WDYT?

pkg/profile/artifact_single_repo_test.go
aclevername: I think thats a smell :wink: I think I would naturally look for this test in `artifact_test.go` since its testing `artifact.go` logic. I think we can do a lot of other work to make the `artifact_tests` smaller https://github.com/weaveworks/pctl/pull/144#discussion_r653552254

pkg/profile/artifact_test.go
Skarlso: I.... I think so? But not the file generation which we also want to test, right? So I'm super on the fence how to handle this. :D

pkg/profile/chart_builder.go
Skarlso: I don't know. :D It's a full copy of the original builder code. :D I'll take a closer look. :)

Skarlso: > // This is an optional value. If defined, it takes precedence over other Chart fields.
This is the Path. So I think, it doesn't matter?

pkg/profile/artifact_test.go
```diff
@@ -157,6 +65,81 @@ var _ = Describe("Profile", func() {
...
+				SparseFolder: "weaveworks-nginx",
+				Branch:       "",
+			},
+		}
+		profile.SetProfileMakeArtifacts(func(pam *profile.ProfilesArtifactsMaker, installation profilesv1.ProfileInstallation) ([]artifact.Artifact, error) {
+			return artifacts, nil
+		})
+	})
+	AfterEach(func() {
+		Expect(os.RemoveAll(rootDir)).To(Succeed())
```
aclevername, line 142: duplicated

pkg/profile/artifact_test.go
```diff
@@ -157,6 +65,81 @@ var _ = Describe("Profile", func() {
 		var err error
 		rootDir, err = ioutil.TempDir("", "test_make_artifacts")
 		Expect(err).NotTo(HaveOccurred())
+		gitRepositoryName = "git-repo-name"
+		gitRepositoryNamespace = "git-repo-namespace"
+		artifacts := []artifact.Artifact{
```
aclevername, line 70: do we need two resources here? could we get away with just 1 for slimming the test down? Or is it testing some functionality differently

aclevername (approved):
LGTM, couple comments :smile:

pkg/profile/artifact_test.go
Skarlso: I guess we don't NEED two. :) But it's nice to have more than one if you have a loop. To make sure that the system behaves consistently on each cycle. That there aren't any things that it reuses incorrectly, or caches, or keeps something that its not supposed to.

PR #145, opened 2021-06-17
Show all catalog entires

Himangini:
### Description
Added --all option for search cmd

### Checklist

### DIFF
--- a/cmd/pctl/main.go
+++ b/cmd/pctl/main.go
@@ -93,6 +93,21 @@ func parseArgs(c *cli.Context) (string, *client.Client, error) {
 	return c.Args().First(), client, nil
 }
 
+func getClient(c *cli.Context) (*client.Client, error) {
+	options := client.ServiceOptions{
+		KubeconfigPath: c.String("kubeconfig"),
+		Namespace:      c.String("catalog-service-namespace"),
+		ServiceName:    c.String("catalog-service-name"),
+		ServicePort:    c.String("catalog-service-port"),
+	}
+
+	client, err := client.NewFromOptions(options)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
 func buildK8sClient(kubeconfig string) (runtimeclient.Client, error) {
 	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
 	if err != nil {
--- a/cmd/pctl/main.go
+++ b/cmd/pctl/main.go
@@ -101,11 +101,7 @@ func getClient(c *cli.Context) (*client.Client, error) {
 		ServicePort:    c.String("catalog-service-port"),
 	}
 
-	client, err := client.NewFromOptions(options)
-	if err != nil {
-		return nil, err
-	}
-	return client, nil
+	return client.NewFromOptions(options)
 }
 
 func buildK8sClient(kubeconfig string) (runtimeclient.Client, error) {
--- a/cmd/pctl/main.go
+++ b/cmd/pctl/main.go
@@ -76,31 +76,30 @@ func globalFlags() []cli.Flag {
 }
 
 func parseArgs(c *cli.Context) (string, *client.Client, error) {
-	options := client.ServiceOptions{
-		KubeconfigPath: c.String("kubeconfig"),
-		Namespace:      c.String("catalog-service-namespace"),
-		ServiceName:    c.String("catalog-service-name"),
-		ServicePort:    c.String("catalog-service-port"),
-	}
-
 	if c.Args().Len() < 1 {
 		return "", nil, fmt.Errorf("argument must be provided")
 	}
-	client, err := client.NewFromOptions(options)
+	client, err := buildCatalogClient(c)
 	if err != nil {
 		return "", nil, err
 	}
 	return c.Args().First(), client, nil
 }
 
-func getClient(c *cli.Context) (*client.Client, error) {
+func getCatalogClient(c *cli.Context) (*client.Client, error) {
+	if c.Args().Len() > 0 {
+		return nil, fmt.Errorf("argument must not be provided")
+	}
+	return buildCatalogClient(c)
+}
+
+func buildCatalogClient(c *cli.Context) (*client.Client, error) {
 	options := client.ServiceOptions{
 		KubeconfigPath: c.String("kubeconfig"),
 		Namespace:      c.String("catalog-service-namespace"),
 		ServiceName:    c.String("catalog-service-name"),
 		ServicePort:    c.String("catalog-service-port"),
 	}
-
 	return client.NewFromOptions(options)
 }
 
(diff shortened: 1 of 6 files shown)


cmd/pctl/main.go
```diff
@@ -93,6 +93,21 @@ func parseArgs(c *cli.Context) (string, *client.Client, error) {
...
+		ServiceName:    c.String("catalog-service-name"),
+		ServicePort:    c.String("catalog-service-port"),
+	}
+
+	client, err := client.NewFromOptions(options)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
```
Callisto13, line 109: can we shorten this to 
```suggestion
	return client.NewFromOptions(options)
}
```

cmd/pctl/search.go
```diff
@@ -25,41 +25,43 @@ func searchCmd() *cli.Command {
...
-				_ = cli.ShowCommandHelp(c, "search")
-				return err
-			}
-
-			profiles, err := catalog.Search(catalogClient, searchName)
-			if err != nil {
-				return err
-			}
+			var profiles []profilesv1.ProfileCatalogEntry
 			outFormat := c.String("output")
```
Callisto13, line 36: can we move this closer to where it is used?

cmd/pctl/search.go
```diff
@@ -25,41 +25,43 @@ func searchCmd() *cli.Command {
...
-			out, err := f.Format(getter)
-			if err != nil {
-				return err
+			outErr := outputFormatter(profiles, outFormat)
+			if outErr != nil {
+				return outErr
 			}
 
-			fmt.Println(out)
 			return nil
```
Callisto13, line 65: I think we can just `return outputFormatter(profiles, outFormat)` here too.

Can we rename this func too? With `outputFormatter` I feel like I should be getting an object back. What do you think of just `formatOutput` or `printOutput`?

cmd/pctl/search.go
Himangini: its used in the following if-else loop. i have moved the `outFormat` var down the line where its used

cmd/pctl/main.go
Himangini: done 👍🏻

cmd/pctl/search.go
Himangini: done , renamed it to `formatOutput` 👍🏻

cmd/pctl/search.go
```diff
@@ -25,42 +25,38 @@ func searchCmd() *cli.Command {
...
+				if err != nil {
+					_ = cli.ShowCommandHelp(c, "search")
+					return err
+				}
+				profiles, err = catalog.Search(catalogClient, "all")
+				if err != nil {
+					return err
+				}
+			} else {
+				searchName, catalogClient, err := parseArgs(c)
```
aclevername, line 47: should we error if `--all` and an arg is provided?

pkg/catalog/search.go
```diff
@@ -9,11 +9,21 @@ import (
...
 func Search(catalogClient CatalogClient, searchName string) ([]profilesv1.ProfileCatalogEntry, error) {
-	q := map[string]string{
-		"name": searchName,
+	var data []byte
+	var statusCode int
+	var err error
+
+	if searchName == "all" {
+		data, statusCode, err = catalogClient.DoRequest("/profiles", nil)
+	} else {
```
aclevername, line 20: does this mean `pctl search all` is equivalent to `pctl search --all`? Also it blocks users from searching for the string `all`, which might be fine if its intentional :smile:

cmd/pctl/search.go
Himangini: i am not sure , what do u think? atm i think if u specify `-all` and searchname it'll ignore the searchname and return all available profiles

pkg/catalog/search.go
Himangini: i think this was not intentional, i'm gonna remove this. nice catch 👌🏻

cmd/pctl/search.go
aclevername: I think its worth erroring if both are set, to avoid confusion. People might think `--all` does something else otherwise

Makefile
```diff
@@ -26,7 +26,7 @@ local-env: submodule ## Create local environment
 
 submodule: ## Update git submodules
 	git submodule init
-	git submodule update
+	git submodule update --remote --merge
```
Skarlso, line 29: This might mess things up when there is a zero checkout. The init and update call are fine if there is no module initially, which is the case on CI when this is running. If you want to update, that's when we force a remote and a merge.

Himangini: i had to do this to get the local tests working , i am gonna put it back to what it was before

Skarlso: Hmm. That is interesting. Maybe leave it in and we'll see?

Skarlso: Oh! I remember why it's that. Because this might cause an involuntary update to running tests locally. Normally, you don't want to force update.

Himangini: i have removed it from the makefile still getting same lint error 😑

cmd/pctl/search.go
Himangini: i have added the error if both `-all` and searchname is provided

cmd/pctl/main.go
```diff
@@ -93,6 +93,21 @@ func parseArgs(c *cli.Context) (string, *client.Client, error) {
...
+		ServiceName:    c.String("catalog-service-name"),
+		ServicePort:    c.String("catalog-service-port"),
+	}
+
+	if c.Args().Len() > 0 {
+		return nil, fmt.Errorf("argument must not be provided")
+	}
+
+	return client.NewFromOptions(options)
+}
```
aclevername, line 109: is this a repeat of the above? It could be re-used instead of duplicated :smile:

aclevername (approved):
1 comment otherwise LGTM :smile:

Skarlso:
I have no idea why the linter is broken. :O

cmd/pctl/main.go
Himangini: yeah I was thinking of refactoring it but got sidetracked by the failing build 😓  I'll refactor it 👍🏻

Himangini: done 👍🏻

cmd/pctl/search.go
```diff
@@ -25,42 +25,38 @@ func searchCmd() *cli.Command {
...
+				if err != nil {
+					_ = cli.ShowCommandHelp(c, "search")
+					return err
+				}
+				profiles, err = catalog.Search(catalogClient, searchName)
+				if err != nil {
+					return err
 				}
 			}
```
Callisto13, line 56: could this be simplified and dried a bit? (not tested, btw, could be wrong)
```suggestion
      searchName, catalogClient, err := parseArgs(c)
      if err != nil {
        _ = cli.ShowCommandHelp(c, "search")
        return err
      }
      if c.Bool("all") {
        searchName = ""
      }
      profiles, err = catalog.Search(catalogClient, searchName)
      if err != nil {
        return err
      }
```

Callisto13 (approved):
lgtm 👍 

could you rebase your commit messages before you merge so that their context is clear?

Issue #134, opened 2021-06-14
Sparse clone clones for each artifact. It should be more clever than that


Skarlso:
After the refactor is done, have a think about how to better do the sparse cloning. If there are multiple artifacts for a single repo, it sparse clones on each artifact, instead of doing it once, then copying out the needed artifact details.

Skarlso:
Local artifact: anything that lives in the profile definition's repository. Things like, deployment.yaml, chart.yaml...

For example, consider the following profiles.yaml file:

```yaml
apiVersion: weave.works/v1alpha1
kind: ProfileDefinition
metadata:
  name: weaveworks-nginx
spec:
  name: weaveworks-nginx
  description: Profile for deploying nginx
  maintainer: weaveworks
  prerequisites:
    - "kubernetes 1.19"
  artifacts:
    - name: nginx-deployment-1
      kustomize:
        path: nginx/deployment-1
    - name: nginx-deployment-2
      kustomize:
        path: nginx/deployment-2
```

These two artifacts are in the same repository. This should result in a single clone and thing to copy twice.

This code lives here: https://github.com/weaveworks/pctl/blob/7b8b82f7220d03f52ed0efa4b4cd4bb068fcde42/pkg/profile/artifact.go#L98

This function works through pathToCopy. But the problem is, that the Artifact is per artifacts instead of per ProfileDefinition.
Maybe have a higher level abstraction?

PR #183, opened 2021-06-28
Sparse clone only per artifact

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/134

Before:

```
time ./pctl install --namespace default --out install --profile-branch main --profile-url https://github.com/weaveworks/profiles-examples --profile-path weaveworks-nginx --git-repository cool-namespace/git-repo
generating installation for profile weaveworks-nginx/main:

./pctl install --namespace default --out install --profile-branch main        0.46s user 0.42s system 11% cpu 7.507 total
```

After:

```
time ./pctl install --namespace default --out install --profile-branch main --profile-url https://github.com/weaveworks/profiles-examples --profile-path weaveworks-nginx --git-repository cool-namespace/git-repo
generating installation for profile weaveworks-nginx/main:

./pctl install --namespace default --out install --profile-branch main        0.11s user 0.14s system 11% cpu 2.216 total
```

That's a significant improvement. From 7 seconds down to 2. :) That's with a repo which has a nested profile in it on a different branch. So it had to be cloned twice.

It gets even better if it's only the bitnami nginx:

Before:
```
time ./pctl install --namespace default --out install --profile-branch main --profile-url https://github.com/weaveworks/profiles-examples --profile-path bitnami-nginx --git-repository cool-namespace/git-repo
generating installation for profile bitnami-nginx/main:

./pctl install --namespace default --out install --profile-branch main        0.25s user 0.24s system 5% cpu 8.821 total
```

After:
```
time ./pctl install --namespace default --out install --profile-branch main --profile-url https://github.com/weaveworks/profiles-examples --profile-path bitnami-nginx --git-repository cool-namespace/git-repo
generating installation for profile bitnami-nginx/main:

./pctl install --namespace default --out install --profile-branch main        0.07s user 0.08s system 12% cpu 1.311 total
```

### Checklist


pkg/profile/artifact_test.go
```diff
@@ -413,6 +395,156 @@ status: {}
 `))
 			})
 		})
+
+		When("there are multiple artifacts for the same repository", func() {
```
aclevername, line 399:
```suggestion
		When("there are multiple artifacts for the same repository and branch", func() {
```

pkg/profile/artifact_test.go
```diff
@@ -413,6 +395,156 @@ status: {}
...
+					RootDir:          rootDir,
+					GitRepoNamespace: gitRepoNamespace,
+					GitRepoName:      gitRepoName,
+				})
+				err := maker.Make(pSub)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(fakeGitClient.CloneCallCount()).To(Equal(2))
+			})
+			It("should still clone same urls but with different branch", func() {
+				fakeGitClient.CloneStub = func(url string, branch string, dir string) error {
```
aclevername, line 527: can we not make this smarter, just checkout branch if repo is already cloned? or is that not really any faster?

pkg/profile/profile.go
```diff
@@ -1,13 +1,7 @@
...
-
-// ProfileGetter is a func that can fetch a profile definition
-type ProfileGetter func(repoURL, branch, path string, gitClient git.Git) (profilesv1.ProfileDefinition, error)
-
-var getProfileDefinition = repo.GetProfileDefinition
+//
+//// ProfileGetter is a func that can fetch a profile definition
+//type ProfileGetter func(repoURL, branch, path string, gitClient git.Git, pam *ProfilesArtifactsMaker) (profilesv1.ProfileDefinition, error)
+//
+//var getProfileDefinition = GetProfileDefinition
```
aclevername, line 7: can be deleted?

pkg/profile/artifact_test.go
Skarlso: That would increase the complexity of this very simple thing. I traded that for the fact that I don't think it happens often enough. WDYT?

pkg/profile/profile.go
Skarlso: Oh ops, thanks.

pkg/profile/repo_test.go
```diff
@@ -0,0 +1,209 @@
...
+		When("called multiple times with different url", func() {
+			It("it clones all", func() {
+				_, err := maker.GetProfileDefinition(repoURL, branch, "my-profile")
+				Expect(err).NotTo(HaveOccurred())
+				_, err = maker.GetProfileDefinition("https://github.com/different", branch, "my-profile")
+				Expect(err).NotTo(HaveOccurred())
+				Expect(fakeGitClient.CloneCallCount()).To(Equal(2))
+			})
+		})
+		When("called multiple times with the same url but different branch", func() {
```
aclevername, line 199: these tests exist in `repo_test` and `artifact_test`?

Skarlso: Yes, because different code path. One is the local checkout and the other is in the profiles cloner.

pkg/profile/artifact_test.go
Skarlso: I mean, not by that much, but I would have to also keep track of what branch is checked out, and then check out the right branch that is needed. Or always checkout the branch when I'm in the right place. It just adds a bit of more logic.

Skarlso: Also, then it's no longer a shallow clone. Because I need all branches and all tags.

Skarlso: Unless, I do a fetch first... then checkout. You see how it gets more complex? :D

Skarlso: Also, it actually wouldn't matter in terms of performance, because we shallow clone a specific tag anyways. Which means, we have to check out / fetch that specific tag. It wouldn't be a difference since in both cases it clones the same amount -metadata I guess. Which is only a couple bytes.

Skarlso: So there you have your answer. :D :D I knew I already went through this, but failed to write it down so I had to think it through again :D

aclevername (approved):

PR #213, opened 2021-07-12
Adding dependsOn to same type resources

Skarlso:
### Description

Closes https://github.com/weaveworks/profiles/issues/214

### TODO
- [x] Check the depended upon artifact's type, because it needs to do something differently if it isn't the same type.

### Checklist


pkg/profile/builders/chart/builder.go
```diff
@@ -108,6 +109,22 @@ func (c *Builder) makeHelmReleaseObjects(artifact profilesv1.Artifact, installat
...
 	}
+	spec := helmv2.HelmReleaseSpec{
+		Chart: helmv2.HelmChartTemplate{
+			Spec: helmChartSpec,
+		},
+		ValuesFrom: values,
+	}
+	var d []dependency.CrossNamespaceDependencyReference
+	for _, dep := range dependencies {
+		if dep.Chart != nil {
```
Skarlso, line 120: A follow up Pr will expand this to deal with different type objects.

tests/integration/integration_test.go
```diff
@@ -940,6 +967,136 @@ status: {}
 
 			Expect(podList.Items[0].Spec.Containers[0].Image).To(Equal("docker.io/bitnami/nginx:1.19.10-debian-10-r35"))
 		})
+		When("there is a depending chart", func() {
```
Skarlso, line 970: this only tests charts right now. not kustomize... eventually, I'll expand the profile to be more complex

pkg/profile/builders/builders.go
```diff
@@ -17,5 +17,5 @@ const (
 //go:generate counterfeiter -o fakes/builder_maker.go . Builder
 type Builder interface {
 	// Build a single artifact from a profile artifact and installation.
-	Build(artifact profilesv1.Artifact, installation profilesv1.ProfileInstallation, definition profilesv1.ProfileDefinition) ([]artifact.Artifact, error)
+	Build(artifact profilesv1.Artifact, installation profilesv1.ProfileInstallation, definition profilesv1.ProfileDefinition, dependencies []profilesv1.Artifact) ([]artifact.Artifact, error)
```
Skarlso, line 20: @aclevername @Callisto13 so I didn't find a better solution than to extend the interface for easy handling of the dependencies. I'm not sure I like this. But it might be fine. I don't know. :D WDYT?

I would rather have it on the builder, but we aren't constructing the builders at that point. It makes sense to construct the list in the make artifacts too because that's what has all the artifacts. So this might be fine, but if you have a better idea, feel free to suggest. :)

tests/integration/integration_test.go
```diff
@@ -940,6 +967,136 @@ status: {}
...
+					err := kClient.List(context.Background(), podList, helmReleaseOpts...)
+					Expect(err).NotTo(HaveOccurred())
+					if len(podList.Items) == 0 {
+						return v1.PodPhase("no pods found")
+					}
+					return podList.Items[0].Status.Phase
+				}, 2*time.Minute, 5*time.Second).Should(Equal(v1.PodPhase("Running")))
+				Expect(podList.Items[0].Spec.Containers[0].Image).To(Equal("docker.io/bitnami/nginx:1.21.0-debian-10-r0"))
+			})
+		})
```
aclevername, line 1099: rather than a new test can we just modify the existing test? This files a 1000 lines :sob:

pkg/profile/builders/builders.go
aclevername: do we need an extra argument? I thought perhaps it already had all the information it needed being passed in

Skarlso: It's in the definition. WTF am I even doing.

tests/integration/integration_test.go
Skarlso: Hmm, hmmm maybe. :D

pkg/profile/builders/builders.go
Skarlso: Oh Gods, this hurts so much man. I don't even know. I blame it on my meds.

tests/integration/integration_test.go
Skarlso: If I modify the existing profile like `weaveworks-nginx`, it will trigger update to almost ever test as in add to the output new information and whatnot. The net change would be the same number of lines as with a new test, but would be more confusing.

tests/integration/integration_search_test.go
```diff
@@ -0,0 +1,179 @@
+package integration_test
```
Skarlso, line 1: This is only an extract of search context into a separate file.

tests/integration/integration_prepare_test.go
```diff
@@ -0,0 +1,51 @@
+package integration_test
```
Skarlso, line 1: This is only an extract of prepare context into a separate file.

tests/integration/integration_list_test.go
```diff
@@ -0,0 +1,109 @@
+package integration_test
```
Skarlso, line 1: This is only an extract of list context into a separate file.

tests/integration/integration_docgen_test.go
```diff
@@ -0,0 +1,81 @@
+package integration_test
```
Skarlso, line 1: This is only an extract of docgen context into a separate file.

tests/integration/integration_show_test.go
```diff
@@ -0,0 +1,68 @@
+package integration_test
```
Skarlso, line 1: This is only an extract of show context into a separate file.

Skarlso:
This PR will be completely superseded by a followup PR which changed how we render HelmReleases and basically dependencies between various resources.

PR #220, opened 2021-07-15
Add dependency handling to artifacts

Skarlso:
### Description

Closes https://github.com/weaveworks/profiles/issues/215

_Note_: Since this is an architectural refactor and there was a LOT of stuff moved, deleted, renamed, I suggest checking out this PR and viewing it in its entirety because the diff is horrible to look at.

### TODO

- [x] Update profiles-example
- [x] Update tests

### Checklist

### Docs

A followup PR is coming in Profiles to update the documentation.


tests/integration/integration_install_test.go
```diff
@@ -854,6 +547,7 @@ status: {}
 		})
 
 		It("generates valid artifacts to the local directory", func() {
+			Skip("Ignored for now, might be deleted or refactoring")
```
Skarlso, line 550: @Skarlso this will be refactored and changed in a later PR

tests/integration/integration_install_test.go
```diff
@@ -940,110 +636,139 @@ status: {}
...
-				Expect(err).ToNot(HaveOccurred())
-				Expect(string(output)).To(ContainSubstring("kind: List"))
-			})
-		})
-		When("the provided version is missing", func() {
-			It("will put out an understandable error message", func() {
-				cmd := exec.Command(binaryPath, "prepare", "--version=vnope")
+		When("there is a depending chart", func() {
+			It("generates artifacts which contain a depends on flag", func() {
+				Skip("Ignored for now, will be deleted or refactoring")
```
Skarlso, line 641: @Skarlso this will be refactored and changed in a later PR

Skarlso:
TODOs including updating profiles-examples with a complex depends on structure AND updating the docs, because the generated files look differently now.

tests/integration/integration_install_test.go
Skarlso: Note, the test works, but is right now using a deprecated way of installing things which does fail. But we are re-thinking the whole integration test process, so I won't bother refactoring it right now.

pkg/profile/artifact.go
```diff
@@ -90,26 +80,51 @@ func (pa *ProfilesArtifactsMaker) Make(installation profilesv1.ProfileInstallati
...
+					return fmt.Errorf("failed to create directory")
+				}
+				filename = filepath.Join(subFolder, fmt.Sprintf("%s.%s", name, "yaml"))
+			}
+			if err := pa.generateOutput(filename, obj.Object); err != nil {
 				return err
 			}
 		}
+		// if we have a local resource, write out the kustomization yaml limiting its visibility.
 		if artifact.Kustomize != nil {
```
aclevername, line 100: is it ever not set?

pkg/profile/artifact/artifact.go
```diff
@@ -5,13 +5,29 @@ import (
...
-	RepoURL      string
-	PathsToCopy  []string
-	SparseFolder string
-	Branch       string
+	// Generated Kubernetes resources for this artifact.
+	Objects []Object
+	// Kustomize resource which limits the number of objects scanned by flux.
+	Kustomize *types.Kustomization
+	// KustomizeWrapper limits flux to only pick up a specific resource.
+	KustomizeWrapper *types.Kustomization
```
aclevername, line 25: :ahhh: This naming is confusing, i can't tell which is doing what?

pkg/profile/artifact/artifact.go
```diff
@@ -5,13 +5,29 @@ import (
...
+	Name             string
+	RepoURL          string
+	PathsToCopy      []string
+	SparseFolder     string
+	Branch           string
+	// If set, use this as folder for all artifacts generated from Objects.
+	SubFolder string
```
aclevername, line 32: this struct is very confusing. Can we reduce it down or have some sub-structs to make it clear what values are related to eachother and what they do?

pkg/profile/builder/builder.go
```diff
@@ -0,0 +1,349 @@
...
+}
+
+// ArtifactBuilder will build helm chart resources.
+type ArtifactBuilder struct {
+	Config
+	a *artifact.Artifact
+}
+
+// Build a single artifact from a profile artifact and installation.
+func (c *ArtifactBuilder) Build(partifact profilesv1.Artifact, installation profilesv1.ProfileInstallation, definition profilesv1.ProfileDefinition) ([]artifact.Artifact, error) {
```
aclevername, line 52: feels weird we pass in the artifact and the whole definition, but I guess its unavoidable for this check
```
	for _, dep := range partifact.DependsOn {
		d, ok := c.containsArtifact(dep.Name, definition.Spec.Artifacts)
		if !ok {
			return nil, fmt.Errorf("%s's depending artifact %s not found in the list of artifacts", partifact.Name, dep.Name)
		}
		deps = append(deps, d)
	}
```

aclevername: perhaps we could move this logic outside of the builder?

pkg/profile/builder/builder.go
```diff
@@ -0,0 +1,349 @@
...
+		if !ok {
+			return nil, fmt.Errorf("%s's depending artifact %s not found in the list of artifacts", partifact.Name, dep.Name)
+		}
+		deps = append(deps, d)
+	}
+	c.a = &artifact.Artifact{Name: partifact.Name, KustomizeWrapper: &types.Kustomization{
+		Resources: []string{"kustomize-flux.yaml"},
+	}}
+	if partifact.Chart != nil {
+		if err := c.applyChartDetailsToArtifact(partifact, installation, definition, deps); err != nil {
```
aclevername, line 65: what does "apply" mean in this context? I don't quite understand the name :confused:

pkg/profile/artifact.go
Skarlso: Yes. This probably comes from the fact that it's named poorly. :D I'll try and rename it to give it better context.

pkg/profile/builder/builder.go
Skarlso: I think I can perform it in the Make phase. I'll take a look.

pkg/profile/builder/builder.go
Skarlso: Hm, I was trying to go for a name which conveys that the inner artifact value will be updated. Maybe updateArtifactWithChartDetails?

aclevername: sounds good!

pkg/profile/artifact.go
```diff
@@ -90,26 +80,51 @@ func (pa *ProfilesArtifactsMaker) Make(installation profilesv1.ProfileInstallati
...
+// writeOutKustomizeResource writes out kustomization resource data if set to a specific file.
+func writeOutKustomizeResource(kustomize *types.Kustomization, filename string) error {
+	data, err := yaml.Marshal(kustomize)
+	if err != nil {
+		return fmt.Errorf("failed to marshal kustomize resource: %w", err)
+	}
+	err = os.WriteFile(filename, data, 0644)
+	if err != nil {
+		return fmt.Errorf("failed to write file %s: %w", filename, err)
+	}
```
Callisto13, line 124: ahem

image

Skarlso: To be fair, that function was an extract of code from something Jake wrote. :D :D :D :D

But superb catch right there in my face. :D

aclevername: WOW BLAME ME HUH :joy: :heart: 

Its your name attached to the git blame once this is merged :wink:

pkg/profile/builder/builder.go
Skarlso: ah! I know why I'm doing this here. :) I'm doing this here, because we need the dependencies and checking them outside, in Make, I would still have to pass them in, which would mean, that I have to modify the Build interface, which don't want. And as you saw, since this has to be performed for each artifact, and various other things, I have to also have the definition. :/

tests/integration/integration_install_test.go
```diff
@@ -5,369 +5,28 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/client"
-
-	kustomizev1 "github.com/fluxcd/kustomize-controller/api/v1beta1"
-	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
-	v1 "k8s.io/api/core/v1"
 )
 
 const (
-	profileExamplesURL                = "https://github.com/weaveworks/profiles-examples"
+	profileExamplesURL                = "https://github.com/Skarlso/profiles-examples"
```
Skarlso, line 25: @Skarlso remember to unpin this once all tests are passing with the new profiles.

tests/integration/integration_install_test.go
```diff
@@ -491,16 +167,16 @@ var _ = Describe("PCTL", func() {
...
 metadata:
   creationTimestamp: null
-  name: pctl-profile
+  name: pprof
   namespace: %s
 spec:
   configMap: %s
   source:
-    branch: main
+    branch: dep-with-wrap
```
Skarlso, line 175: @Skarlso Remember to unpin this.

pkg/profile/artifact/artifact.go
```diff
@@ -5,13 +5,33 @@ import (
...
+}
+
+// Kustomize contains resources which direct flux to what to install.
+type Kustomize struct {
+	// LocalResourceLimiter resource which limits the number of objects scanned by flux.
+	LocalResourceLimiter *types.Kustomization
+	// ObjectWrapper is the top most Kubernetes object wrapper for all resources.
+	ObjectWrapper *types.Kustomization
+}
+
```
aclevername, line 25: I still find this really confusing :sweat_smile: but I'm honestly struggling to think of something better. Short of calling them `KubernetesKustomize` and `FluxKustomize` :joy:

pkg/profile/builder/builder.go
```diff
@@ -0,0 +1,351 @@
...
+}
+
+// ArtifactBuilder will build helm chart resources.
+type ArtifactBuilder struct {
+	Config
+	a *artifact.Artifact
+}
+
+// Build a single artifact from a profile artifact and installation.
+func (c *ArtifactBuilder) Build(partifact profilesv1.Artifact, installation profilesv1.ProfileInstallation, definition profilesv1.ProfileDefinition) ([]artifact.Artifact, error) {
```
aclevername, line 52: do we wanna add a 
```
var _ Builder := &ArtiactBuilder{}
```

I saw you do that elsewhere for checking interface implementation and I liked it :smile:

pkg/profile/builder/builder.go
```diff
@@ -0,0 +1,351 @@
...
+	var deps []profilesv1.Artifact
+	for _, dep := range partifact.DependsOn {
+		d, ok := c.containsArtifact(dep.Name, definition.Spec.Artifacts)
+		if !ok {
+			return nil, fmt.Errorf("%s's depending artifact %s not found in the list of artifacts", partifact.Name, dep.Name)
+		}
+		deps = append(deps, d)
+	}
```
aclevername, line 60: do we need to check if its depending on a nested profile? or I guess we can skip it since its the next story coming up

pkg/profile/builder/builder.go
```diff
@@ -0,0 +1,351 @@
...
+type Config struct {
+	GitRepositoryName      string
+	GitRepositoryNamespace string
+	RootDir                string
+}
+
+// ArtifactBuilder will build helm chart resources.
+type ArtifactBuilder struct {
+	Config
+	a *artifact.Artifact
```
aclevername, line 48: Can we rename this to artifact? since its on the struct it won't collide with the package import

pkg/profile/builder/builder.go
Skarlso: :D :D noicee :) Yeah, I'll add it ;)

pkg/profile/builder/builder.go
Skarlso: Yeah thought about that too. And yes, I think the follow up PR will "fix" that. As in it will be a valid choice?

pkg/profile/artifact/artifact.go
Skarlso: The problem with that is that both are there because of Flux. :D

aclevername (approved):
LGTM. Still not 100% on the kustomize naming, but lets run with it for now and we can always refactor it :smile:

PR #235, opened 2021-07-22
Add support for depending on a nested profile

aclevername:
Adds support for setting `dependsOn` against a nested profile. Majority of this PR is refactoring the `pkg/install` code so that it collects all the artifacts (including ones from nested) and then invokes the artifactWriter to writer all of the flux resources and copy the files across.

### Checklist


tests/integration/integration_install_test.go
```diff
@@ -71,7 +72,8 @@ var _ = Describe("pctl install", func() {
 				Skip("Skipping this tests as it requires credentials")
 			}
 
-			profileBranch := "main"
+			//TODO change backt to main
+			profileBranch := "depends-nested"
```
aclevername, line 76: TODO revert and merge https://github.com/weaveworks/profiles-examples/pull/12

pkg/install/artifact/artifact.go
```diff
@@ -1,37 +1,408 @@
...
+
+// ArtifactWriter can build an artifacts from an installation and a profile artifact.
+//go:generate counterfeiter -o fakes/fake_builder.go . ArtifactWriter
+type ArtifactWriter interface {
+	Write(installation profilesv1.ProfileInstallation, artifacts []ArtifactWrapper) error
+}
+
+var _ ArtifactWriter = &Writer{}
+
+type ArtifactWrapper struct {
```
Skarlso, line 41: Missing comment. :) And there are some other Exported methods and structus which are missing comments. I won't tag them all. :D

pkg/install/installer.go
```diff
@@ -35,178 +30,127 @@ type Config struct {
...
+				PathToProfileClone: filepath.Join(i.clonedRepos[profileRepoKey], installation.Spec.Source.Path),
+				ProfileName:        profileDef.Name,
 			}
-		}
-		filename := filepath.Join(artifactDir, "kustomization.yaml")
-		if err := writeOutKustomizeResource(artifact.Kustomize.ObjectWrapper, filename); err != nil {
-			return err
+			//if its a nested artifact we want to wrap the artifact in a sub-directory
+			if isNestedArtifact {
+				newArtifact.NestedProfileSubDirectoryName = installation.Name
```
Skarlso, line 106: This should do something like filepath.Join. Right now it doesn't nest double nested profiles. Structures like this:

```
nested artifacts A -> nested artifacts B -> nginx
```

Ends up in something like this:

```
nested artifacts from A
nested artifacts from B
```

This will cause a problem if the artifacts happened to be named the same. Then they will share a folder and overwrite each other.

Skarlso: I created a test profile to test that here: https://github.com/weaveworks/profiles-examples/tree/double_nested

Skarlso (commented):
Well done! One remark regarding deeper nested profiles and some notes on missing comments.

But otherwise, I like this refactor. I would have further factored out the `make*` functions, that's why there was a builder after all, however, I can see how that would duplicate a lot of the tests that we perform. I guess that's why you removed it?

Because it's funny how, the `writer` also `makes` if you see what I mean? :)

Great job, nice simplification of the artifacts being written out into a flat structure. 👍

pkg/install/installer.go
aclevername: mhmm good point! will update and add a test

aclevername: Added unit test for scenario, which appears to have fixed it :smile: 

Also just for sanity manaully tried our your branch, LGTM:
```
├── artifacts
│   ├── dependon-chart
│   │   ├── helm-chart
│   │   │   ├── ConfigMap.yaml
│   │   │   ├── HelmRelease.yaml
│   │   │   └── HelmRepository.yaml
│   │   ├── kustomization.yaml
│   │   └── kustomize-flux.yaml
│   ├── nested-nginx
│   │   ├── nested-profile2
│   │   │   └── nginx-server
│   │   │       ├── helm-chart
│   │   │       │   ├── HelmRelease.yaml
│   │   │       │   ├── kustomization.yaml
```

Skarlso (approved):

pkg/install/installer.go
Skarlso: Screenshot 2021-07-27 at 8 38 35

I had this before your change. :) Now it does work, yes. :D

Skarlso: Also, I expect there is an nginx-server besides nested profile2, right? Which is the local resource of the nested-nginx thing. :)

Skarlso: Screenshot 2021-07-27 at 8 40 18

Basically, this is now as I would expect it. With two nested profiles each in their respective folder plus, the local resource.

aclevername: I don't quite follow, is the current behaviour correct :see_no_evil: ?

PR #244, opened 2021-07-29
Add support for bootstraping repo with pctl config

aclevername:
### Description

Add a `pctl bootstrap` command to populate your git repository with a `.pctl/config.yaml`. This config file holds the git repository details, removing the need for the user to provide `--git-repository` in further `pctl add` commands.

```
$ pctl bootstrap --git-repostory foo/bar
bootstrap completed

$ cat .pctl/config.yaml
gitRepository:
  name: bar
  namespace: baz
```

### Checklist


pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,69 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/weaveworks/pctl/pkg/runner"
+	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
+	"gopkg.in/yaml.v2"
+)
+
+type Config struct {
```
Skarlso, line 14: Missing comment. :D :D

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,69 @@
...
+	// GitRepository is the git repository flux resource the installation uses
+	GitRepository profilesv1.GitRepository `json:"gitRepository,omitempty"`
+}
+
+var r runner.Runner = &runner.CLIRunner{}
+
+func CreateConfig(namespace, name, directory string) error {
```
Skarlso, line 21: Here too :P

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,69 @@
...
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to marshal config: %w", err)
+	}
+
+	return os.WriteFile(filepath.Join(pctlDir, "config.yaml"), data, 0644)
+}
+
+func GetConfig(directory string) (*Config, error) {
```
Skarlso, line 48: And here ofc. :D

pkg/bootstrap/bootstrap.go
Skarlso: And all the rest. :D

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,69 @@
...
+	// GitRepository is the git repository flux resource the installation uses
+	GitRepository profilesv1.GitRepository `json:"gitRepository,omitempty"`
+}
+
+var r runner.Runner = &runner.CLIRunner{}
+
+func CreateConfig(namespace, name, directory string) error {
+	out, err := r.Run("git", "-C", directory, "rev-parse", "--show-toplevel")
+	if err != nil {
+		if strings.Contains(string(out), "not a git repository") {
```
Skarlso, line 24: We could check this ourselves earlier when the command is called by simply doing an os.Stat on `.git` folder.

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,69 @@
...
+			return fmt.Errorf("the target directory %q is not a git repository", directory)
+		}
+		return fmt.Errorf("failed to get git directory location: %w", err)
+	}
+
+	pctlDir := filepath.Join(strings.TrimSuffix(string(out), "\n"), ".pctl")
```
Skarlso, line 30: If we don't include the folder ( why the folder? are we planning on storing something else in it other than the config? ), we could name this `.pctl.config.yaml`, then this whole upper part of the function is the same as below in the GetConfig and could be extracted into a common function.

Even so, there is some small amount of duplication in here, but I think that's fine.

pkg/bootstrap/bootstrap.go
Skarlso: We aren't modifying Config and you aren't checking nil on it either, so there is no need to share Config up the stack. Remove the pointer, it doesn't need to be one.

pkg/bootstrap/bootstrap.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	configPath := filepath.Join(strings.TrimSuffix(string(out), "\n"), ".pctl", "config.yaml")
+
+	out, err = os.ReadFile(configPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read config file: %w", err)
+	}
+
+	config := &Config{}
+	err = yaml.Unmarshal(out, config)
```
Skarlso, line 64: You know it. :DDDDD

pkg/bootstrap/export_test.go
```diff
@@ -0,0 +1,7 @@
+package bootstrap
+
+import "github.com/weaveworks/pctl/pkg/runner"
+
+func SetRunner(r2 runner.Runner) {
```
Skarlso, line 5: Grumble grumble, I'm not sure I like this. Even package local values should generally be avoided. But I don't have a strong opinion on it.

pkg/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,138 @@
...
+		})
+
+		When("the directory is not a git directory", func() {
+			It("returns an error", func() {
+				err := bootstrap.CreateConfig("foo", "bar", temp)
+				Expect(err).To(MatchError(fmt.Sprintf("the target directory %q is not a git repository", temp)))
+			})
+		})
+
+		When("the directory is not a git directory", func() {
```
Skarlso, line 63: You have the title twice, above as well.

pkg/bootstrap/bootstrap_test.go
```diff
@@ -0,0 +1,138 @@
+package bootstrap_test
```
Skarlso, line 1: Bonus cookie if you cover 

```go
return nil, fmt.Errorf("failed to get git dir location: %w", err)
```

In GetConfig. :D

Skarlso (approved):
Some small notes, missing comments, and general advice. Structure is good!

Well done!

👾

pkg/bootstrap/bootstrap.go
aclevername: its a pointer because the config might not exist, and no error occurs, so we need a way of returning "I didn't error, but the config doesn't exist", so having it be nil is the simplest solution.

pkg/bootstrap/bootstrap.go
aclevername: this approach it also finds the path to the top-level directory for you. You can run bootstrap from within the a sub directory and it will still work. I think its nice/cleaner than looking for `.git` dir

pkg/bootstrap/bootstrap.go
Skarlso: Sure, but you aren't checking if it's nil then.

```go
		config, err := bootstrap.GetConfig(wd)
		if err == nil {
			gitRepoNamespace = config.GitRepository.Namespace
			gitRepoName = config.GitRepository.Name
		}
```

Which means it will error out.

Also, every path you have in GetConfig returns with an error if the file can't be accessed.

Skarlso: And now that I think about it, the error is disregarded, right? We don't log it in any format anywhere. Or am I missing something here / just can't read? :)

aclevername: oops this is missing haha. should be `if err == nil && config != nil`. Because configs are optional, I wasn't really sure how to handle the error. Ideally we'd have a debug log level to print too, but we don't 😕

pkg/bootstrap/export_test.go
aclevername: :shrug:

pkg/bootstrap/bootstrap.go
Skarlso: On second thought, I'm okay with the folder.

aclevername: >  why the folder? are we planning on storing something else in it other than the config

Potentially yes. Mainly just following the `.wego/` convention. Happy to change it if you feel strongly :smile:

PR #271, opened 2021-08-10
Shorten generated resources names 

### LINKED ISSUE #212: Some of the names that we generate for pctl artifacts are getting too long
```
pctl-profile-depends-on-nginx-depend-on-this-redis-master-85744d7479
```

Is longer than the allowed 63 characters.

```
-master-85744d7479
```

Is added by redis, but the rest is us.

### Labels
Lets shorten the name to `<profile-installation-name>-<sha>` (look at how pods get a sha added at the end) and set labels with the fully descriptive names. Labels for each component of the name, e.g.:
- profile name
- branch name/tag
- artifact name
- nested profile

### installation.yaml
Keep our current naming, save it in installation.yaml, read it back when needed

### Make it shorter
Come up with a naming schema which is shorter while staying unique and useful

aclevername:
### Description

Closes #212 

shorten artifact names to be <installation-name>-<artifact-name>

# Before 
 `pctl add --name jake-rocks nginx-catalog/weaveworks-nginx/v0.1.1` would generated:
```
Kustomizations:
- jake-rocks-bitnami-nginx-nginx-server
- jake-rocks-weaveworks-nginx-nginx-chart
- jake-rocks-weaveworks-nginx-nginx-deployment
HelmReleases: 
- jake-rocks-bitnami-nginx-nginx-server
- jake-rocks-weaveworks-nginx-nginx-chart
HelmRepository:
- jake-rocks-profiles-examples-nginx
Pods:
- jake-rocks-bitnami-nginx-nginx-server-cd77dfddc-sjqgx
- jake-rocks-weaveworks-nginx-nginx-chart-7dc8fdb869-dmd6b
- nginx-deployment-66b6c48dd5-8ch7k
```

# After:
```
Kustomizations:
- jake-rocks-nginx-server
- jake-rocks-nginx-chart
- jake-rocks-nginx-deployment
HelmReleases: 
- jake-rocks-nginx-server
- jake-rocks-nginx-chart
HelmRepository:
- jake-rocks-nginx-chart
Pods:
- jake-rocks-nginx-server-56b64df496-fdqn7
- jake-rocks-nginx-chart-78945789fb-x264r
- nginx-deployment-66b6c48dd5-vtn49
```

In order to further reduce the names of the produced pods, we can set a [releaseName](https://fluxcd.io/docs/components/helm/helmreleases/#specification) in the HelmRelease spec, this allows you to shorten the name of what is produced by the HelmRelease, e.g. the pods. The way `releaseName` works is a bit odd. I've observed the following behaviour:
When deploying nginx via a `HelmRelease` 
- Setting `releaseName: foo` results in pods with `foo-nginx-<hashes>`. The chart name `nginx` was appended after `foo`
- Setting `releaseName: nginx-foo` results in pods with `nginx-foo-<hashes>`. The chart name was not appended.

The behaviour is a bit strange, and might mean that if an artifact name contains the name of its helm chart, it will have a different naming convention to one that does not. However it does provide a useful way to shorten the name further, producing:

```
Kustomizations:
- jake-rocks-nginx-server
- jake-rocks-nginx-chart
- jake-rocks-nginx-deployment
HelmReleases: 
- jake-rocks-nginx-server
- jake-rocks-nginx-chart
HelmRepository:
- jake-rocks-nginx-chart
Pods:
- nginx-server-56b64df496-fdqn7
- nginx-chart-78945789fb-x264r
- nginx-deployment-66b6c48dd5-vtn49
```

I'm not sure how keen I am on keeping the `releaseName` setting. I've done that in a separate commit, so we can decide now if we want to keep it or not.

### Checklist


pkg/install/artifact/artifact.go
```diff
@@ -267,10 +268,10 @@ func (c *Writer) makeHelmReleaseObjects(artifact profilesv1.Artifact, installati
 	return helmRelease, cfgMap
 }
 
-func (c *Writer) makeHelmRepository(url string, name string, installation profilesv1.ProfileInstallation) *sourcev1.HelmRepository {
+func (c *Writer) makeHelmRepository(url, artifactName string, installation profilesv1.ProfileInstallation) *sourcev1.HelmRepository {
```
Himangini, line 271: missing type for url? 🤔

aclevername: this is a golang feature :smile: Rather than:
```
func(foo string, bar string)
```

you can do:
```
func(foo, bar string)
```

Himangini: wow, I had no clue you could do that with go, I learnt something new yay 👍🏻

Callisto13:
🤔  the `releaseName` feature is interesting and I can't decide whether slightly different naming conventions are a problem....


I guess if I were a user, and I was trying to debug where a profile install went wrong, I may get confused by the different patterns. You know how we look for common patterns which let us guess "oh this is named like that, so they must be connected" or " i will look out for stuff which looks like this in future". So I _think_ this may just make it harder for users than it needs to be.

The more important question would be: do we need it? Is the other stuff you have done enough to solve our "names too long" problem, or will it still be a thing and `releaseName` is our only option? If names are short enough then I think we can do without it.

aclevername:
> The more important question would be: do we need it? Is the other stuff you have done enough to solve our "names too long" problem, or will it still be a thing and `releaseName` is our only option? If names are short enough then I think we can do without it.

Yeah this is the choice we have. The last commit was to change to the `releaseName` setting, so it can easily be reverted.


> I guess if I were a user, and I was trying to debug where a profile install went wrong, I may get confused by the different patterns. You know how we look for common patterns which let us guess "oh this is named like that, so they must be connected" or " i will look out for stuff which looks like this in future". So I _think_ this may just make it harder for users than it needs to be.

I don't know for sure, but I'd *imagine* (/hope) that if a HelmRelease deploy failed, inspecting the HelmRelease resources & helm-controller logs would provide the majority of the information to see what went wrong.

Callisto13:
> I don't know for sure, but I'd imagine (/hope) that if a HelmRelease deploy failed, inspecting the HelmRelease resources & helm-controller logs would provide the majority of the information to see what went wrong.

You'd hope. But I can tell you that I am one of those people who will go straight for the pods 😂

Callisto13:
> I don't know for sure, but I'd imagine (/hope) that if a HelmRelease deploy failed, inspecting the HelmRelease resources & helm-controller logs would provide the majority of the information to see what went wrong.

You'd hope. But I can tell you that I am one of those people who will go straight for the pods 😂

Himangini:
I don't mind the releaseName changes, we can try it maybe and wait for some feedback from users?

aclevername:
We could go without the `releaseName` for now, and adjust if people find the names are still too long. Changing it down the line *shouldn't* cause any breaking changes, since its just a modification to the `HelmRelease` spec, not to any of the resources we manage.

aclevername:
> I don't mind the releaseName changes, we can try it maybe and wait for some feedback from users?

Yeah. Its always relatively simple to change, and because its just a change to the `HelmRelease` spec and not the naming of resources we actually create ourselves it shouldn't be a big deal to change down the line.

Himangini (approved):
LGTM 👍🏻

PR #272, opened 2021-08-10
Add tickmarks and flux-like output to pctl output

### LINKED ISSUE #245: Add flux-like tickmarks and output to all output
We should make sure we match the UX style of flux in the output of all commands, for example:
- tickmarks e.g. ✔, ►, etc. ✨
- style for wording
- style for error messages

`pctl add` and `pctl install` in particular should be adjusted (other commands provide a standard `get` output)

From flux:
https://github.com/fluxcd/flux2/blob/4f4f0d70a41e56d8b11e5f4ec1dcd2ae76d89c7f/cmd/flux/log.go#L28-L50

nikimanoledaki:
### Description

Closes #245  

**Implementation details**
- Added a logger that prints to stdout in `pkg/log/log.go`
- Updated the formatting of all commands to add flux-like tickmarks to all logs when generating, erroring, indicating success, and waiting for resources. 

**Demo!** ✨
`pctl install`:
```
./pctl install                 
► checking that flux namespace exists
✔ found flux namespace 'flux-system'
► checking for flux CRDs
✔ found flux CRDs
◎ waiting for resources to be ready
✔ profiles-controller-manager: deployment ready
✔ resources ready
```

### Checklist


pkg/log/log.go
```diff
@@ -0,0 +1,34 @@
...
+
+type StderrLogger struct {
+	Stderr io.Writer
+}
+
+func (l StderrLogger) Actionf(format string, a ...interface{}) {
+	fmt.Fprintln(l.Stderr, `►`, fmt.Sprintf(format, a...))
+}
+
+func (l StderrLogger) Generatef(format string, a ...interface{}) {
```
aclevername, line 16: what does `Generate` mean in the context of logging :thinking: the others I can get onboard with, athough `Action` is also a bit ambiguous. I can't really tell from the name of either of them when they should be used. 

I know we just copied from flux but we can always adjust it :smile:

nikimanoledaki: good point! we only use `Generate` once so it could be replaced by something else, like `Action`. `Action` is used for logs that just announce what's going to happen next but it's very similar to `Waiting`

pkg/log/log.go
```diff
@@ -0,0 +1,31 @@
...
+	format(`⚠️`, m, a...)
+}
+
+func (l PrintLogger) Failuref(m string, a ...interface{}) {
+	format(`✗`, m, a...)
+}
+
+func format(tickmark, m string, a ...interface{}) {
+	fmt.Println(tickmark, fmt.Sprintf(m, a...))
+}
```
aclevername, line 31: do we need to have an instance of the logger? We could change it so they are just public functions instead, like we do in eksctl (using [kris-nova/logger](https://github.com/kris-nova/logger/blob/d961bfacb4a33574805af7cb715567442bdb58d7/logger.go#L109-L168))

nikimanoledaki: you're right, we don't! it was only needed when printing to stdout. thanks, refactoring now :)

pkg/log/log_test.go
```diff
@@ -0,0 +1,36 @@
...
+		tmp = os.Stdout
+		os.Stdout = w
+	})
+
+	It("logs a message to stdout with a tickmark", func() {
+		defer func() {
+			os.Stdout = tmp
+		}()
+
+		log.Warningf("test-warning")
```
nikimanoledaki, line 30: Only testing one of the log funcs because they're all `format()` and only the tickmark changes.

pkg/cluster/install.go
```diff
@@ -122,22 +123,22 @@ func (i *Installer) Install() error {
...
 	}
-	fmt.Println("done.")
-	fmt.Print("Checking for flux CRDs...")
+	log.Successf("found flux namespace '%v'", i.FluxNamespace)
+	log.Actionf("checking for flux CRDs")
 	output, err := i.Runner.Run(kubectlCmd, "get", "crds", "--output", "jsonpath='{.items[*].spec.names.singular}'")
 	if err != nil {
 		if i.IgnorePreflightErrors {
-			fmt.Println("WARNING: failed to list all installed crds. Flux is required for profiles to work.")
+			log.Warningf("WARNING: failed to list all installed crds. Flux is required for profiles to work.")
```
aclevername, line 141: Do we need the "WARNING" text if its got the warning symbol :thinking: ? Maybe it makes it more clear, or perhaps we can move the "WARNING" text into the func itself. Wdyt?

aclevername (approved):
LGTM. 1 comment. I haven't checked but I'm assuming if you grep for `fmt.Print` there are no longer any occurrences as its all just the logger now?

pkg/cluster/install.go
nikimanoledaki: Good idea! Consistency is key, right??!!

nikimanoledaki:
> LGTM. 1 comment. I haven't checked but I'm assuming if you grep for `fmt.Print` there are no longer any occurrences as its all just the logger now?

I'm glad you pointed this out! I found a bunch of `fmt.Print` that I missed 😅  Changes are in https://github.com/weaveworks/pctl/pull/272/commits/ff6cabafe134027c87feffca41e0d3f1a46c1fdb, tests pass

PR #303, opened 2021-08-31
ensure git-repository value is provided in all pctl add scenarios

### LINKED ISSUE #295: Warn the user if the GitRepository reference will be empty
**What were you trying to accomplish?**

Run a pctl add to an environment bootstraped by Weave GitOps. Flux is installed in the wego-system namespace

**What happened?**

The kustomize-flux.yaml referred to a GitRepository with a blank name and no namespace.
```yaml
  sourceRef:
    kind: GitRepository
    name: ""
```
**How to reproduce it?**

don't provide `--git-repository` or a bootstrap config file when running `pctl add`

**Logs**

**Anything else we need to know?**

This isn't major as the workaround is easy enough.  However, if we can warn the user that the profile won't be pulled into the cluster that can save some debugging time.

**Versions**
0.7.0

aclevername:
### Description

Closes #295 

Previously if you installed a profile that only consisted of remote-helm charts we wouldn't error if the user didn't provide the `git-repostory` value, despite needing it.

### Checklist


Skarlso (approved):

cmd/pctl/add.go
```diff
@@ -241,6 +224,27 @@ func addProfile(c *cli.Context) (string, error) {
...
 
+func getGitRepositoryNamespaceAndName(c *cli.Context) (string, string, error) {
+	gitRepository := c.String("git-repository")
+	if gitRepository != "" {
+		split := strings.Split(gitRepository, "/")
+		if len(split) != 2 {
+			return "", "", fmt.Errorf("git-repository must in format <namespace>/<name>; was: %s", gitRepository)
+		}
+		return split[0], split[1], nil
+	} else {
```
Skarlso, line 235: The else is not needed since you are returning inside.

Issue #308, opened 2021-09-08
[Spike] Expose Profiles API


bigkevmcd:
**What feature/behavior/change do you want?**
The profiles "API" should be available for importing into other projects.

**Why do you want this feature?**
As we integrate profiles into other components, we want to be able to use the same logic as `ptcl` uses to manage profiles in repositories.

To do this, we should adopt an architecture that makes this easier, fore example `cmd/pctl/add.go` contains the logic to add profiles to a repository, this should be more easily callable from other projects.

The role of the `cmd/pctl` code should be to get in data, and pass it through to the core logic to execute.

Himangini:
Needs a spike to flesh out the details for this work.

Skarlso:
The attached PR explored refactoring some of things we do in the CLI region. I exposed the aforementioned AddProfile but it turns out that that's pretty much we can sanely expose from what the CLIs are doing and still add some value to them.

The rest of the CLI stuff is using little logic in how in constructs and uses functionality available in the `pkg` section for outside importing. Most of the things it does is basically validation on the parameters and such.

Never the less, I believe that the PR does add some value, if nothing else, I did refactor some pieces of code which got a bit simpler because of that and the `AddProfiles` function IS exported reducing duplication for Team Plutonium.

The architecture I eventually went with is having another package called `api` which holds things that the CLI calls and can be a high level calling from outside of PCTL by other users. I found that this package has some relevance in the Go community as a whole and thus was a good choice to boot. I explored putting some functionality inside `pkg` but that doesn't signify the usage structure as much and looses focal point ( entry point ) to PCTL. While the separate `api` package signifies usage better by saying, hello, import and use me.

Feel free to object / suggest different options. :)

Skarlso:
If the attached PR is acceptable, I'll update it with testing and make it an official change.

Skarlso:
Expanded with unit tests.

Skarlso:
What is the desired outcome of this ticket now? What should the Wego peeps expect?

Skarlso:
So the spike was made, I consider this ticket closed.

PR #311, opened 2021-09-14
Expose api

Skarlso:
### Description

Study and spike on https://github.com/weaveworks/pctl/issues/308

For a full explanation of this PR PLEASE CHECK THE ATTACHED ISSUE. :)

TODO:

- [x] Add tests to the exported AddProfiles function

### Checklist

### DIFF
--- a/api/add.go
+++ b/api/add.go
@@ -3,11 +3,7 @@ package api
 import (
 	"errors"
 	"fmt"
-	"os"
-	"path/filepath"
-	"strings"
 
-	"github.com/weaveworks/pctl/pkg/bootstrap"
 	"github.com/weaveworks/pctl/pkg/catalog"
 	"github.com/weaveworks/pctl/pkg/git"
 	"github.com/weaveworks/pctl/pkg/install"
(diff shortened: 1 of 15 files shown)


Skarlso:
@Callisto13 Added 94.3% line coverage. :D

api/bootstrap.go
```diff
@@ -0,0 +1 @@
+package api
```
Skarlso, line 1: These empty files can be removed if people see fit.

Callisto13:
> @Callisto13 Added 94.3% line coverage. :D

😂  what does this mean?

Skarlso:
> > @Callisto13 Added 94.3% line coverage. :D
> 
> 😂 what does this mean?

Test coverage is 94.3%. :D Run ginkgo with `-cover`. :)

Callisto13:
does test coverage mean anything tho :trollface:

Callisto13:
But this is a spike so wont be merged so 🤷‍♀️

Callisto13:
Okay so what am I reviewing here?

If it is a spike then I will give it a cursory glance and not approve as, by definition, it wont be mergeable. If it is a spike is there a write-up of any learnings? Decisions made and ideas discarded?

If it is actual code we want to keep then I know how much time to spend on it.

Skarlso:
@Callisto13 Uh, that's way too good of a question.

I'll be honest, now that I worked on it for a couple hours trying to come up with the architecture and figure out the testing and the bla bla bla, I'm more inclined to actually lean towards the later. But, then again, there are some pieces in there which are spike related still, things like the empty files for example which only note that things from those commands go in there. BUT I can remove those ofc.

The spike also done as a research so this could be a general outcome of that that is actually usable. What do you think? I lean towards the later, ( especially after adding the tests too ).

Skarlso:
>  If it is a spike is there a write-up of any learnings?

Yes, in the attached ticket. :)

api/add.go
```diff
@@ -0,0 +1,115 @@
...
+	SubName       string
+	Namespace     string
+	ConfigMap     string
+	Dir           string
+	Path          string
+	GitRepository string
+	ProfilePath   string
+	CatalogClient catalog.CatalogClient
+	GitClient     git.Git
+}
```
aclevername, line 30: not all of these options can be configured at the same time right? E.g. ProfilePath vs URL. Is there a way we could convey this through the API?

api/add.go
```diff
@@ -0,0 +1,115 @@
...
+type AddProfileOpts struct {
+	URL           string
+	Branch        string
+	SubName       string
+	Namespace     string
+	ConfigMap     string
+	Dir           string
+	Path          string
+	GitRepository string
+	ProfilePath   string
```
aclevername, line 27: This is the `catalog-name/profile-name` right? Might make sense to have them as separate strings now instead?

api/add.go
```diff
@@ -0,0 +1,115 @@
...
+// AddProfileOpts defines settings for adding profiles.
+type AddProfileOpts struct {
+	URL           string
+	Branch        string
+	SubName       string
+	Namespace     string
+	ConfigMap     string
+	Dir           string
+	Path          string
+	GitRepository string
```
aclevername, line 26: Same here, could be namespace and name separately

api/add.go
Skarlso: If I take all that out into the CLI this looses most of its usefulness. Like, the point of this was that as the data comes in you call it with that and than this just takes care of the rest. That's the most useful part of being in here. If I remove all of those checks, there is nothing useful about this any longer. You could just construct `install` and call that like team wego already does here: https://github.com/weaveworks/weave-gitops-enterprise/pull/50

Skarlso: https://github.com/weaveworks/weave-gitops-enterprise/pull/50/files#diff-24a52d493110426665082003b7a39ca6d6e2b5b75d1a162043f460c238c7a1fbR147

Skarlso: I definitely see value though in adding validation for these values and the API taking care of that part.

api/add.go
Skarlso: Yeah, I see in value in that.

api/add.go
aclevername: > If I take all that out into the CLI this looses most of its usefulness. Like, the point of this was that as the data comes in you call it with that and than this just takes care of the rest. 

but it sort of leaks the fact your using a CLI in the API right? If we imagine a world where this is happening in automation or through a UI this grouping makes less sense

> That's the most useful part of being in here. If I remove all of those checks, there is nothing useful about this any longer. 

I disagree. The value is just having the API, and validating the args.

api/add.go
bigkevmcd: Or even better `NamespacedName` with better typing!

api/add.go
bigkevmcd: I suspect that `Dir` isn't part of this too?

I don't think all these fields are really part of the `AddProfileOpts` ?

Skarlso: The reason they are here, is that without it, there is little value in having this function. I tried to put as much as possible from the CLI into the API. Otherwise, just construct the manager.Install yourself, which wego is already doing. :)

I get how this seems as not really part of an API. But without these things, I don't think there is too much value in extracting AddProfile.

bigkevmcd: I do think there's value, for folks _outside_ of pctl, consider this as an implementation of the Facade pattern?

api/add.go
aclevername: the fact they are together is a technical implementation of our CLI though. In WeGo maybes its `--git-repository-name` and `--git-repository-namespace`, and then they have to concatenate it into our format to use our API?

api/add.go
aclevername: > I suspect that `Dir` isn't part of this too?
> 
> I don't think all these fields are really part of the `AddProfileOpts` ?

do you mean that some are mandatory vs optional?

bigkevmcd: Basically, yes...

In my view, all the exposed functions for the profiles client should take the directory to operate from as the _first_ parameter, there may be other parameters that should be in this list too

api/add.go
Skarlso: Okay, yeah, that's a fair point. :)

api/add.go
Skarlso: Alright, will do it like that. :)

Skarlso: We are going to duplicate some checks in the api and the cli then I guess.

Skarlso:
Wait, we aren't? I thought it will be. :D

aclevername:
> Wait, we aren't? I thought it will be. :D

the story was a spike right? That means an exploration to see what it could be like. If we decide to go down this route we totally can copy stuff, but spike codes are generally intended to be merged

Skarlso:
But the wego folks said that, this looks awesome. :D 

Meh. :) I'm not complaining, just saying that this was something useful for them already.

aclevername:
> But the wego folks said that, this looks awesome. :D
> 
> Meh. :) I'm not complaining, just saying that this was something useful for them already.

right but I'm not disagreeing with its value or saying we won't re-use 99.99% of it. Its just that original story was a spike- an investigation to see what exposing our code as an API could look like. We played the spike before we decided if we were going to go down the root that we need to have an API or not, so the PR was never expected to be merged.

Himangini:
@Skarlso can we close this now? We can revisit this again if we need to in future :)

Issue #309, opened 2021-09-09
As a User I want to be able to be able to configure the default location my Profile installations go into


aclevername:
Currently running `pctl add --name foo my-catalog/weaveworks-nginx` results in an `weaveworks-nginx` directory getting created in the same directory you ran the command in. Users can configure a different location by specifying the output directory, e.g. `pctl add --name foo -o put/it/here/ my-catalog/weaveworks-nginx`. 

Some users might always want the profiles to go into the same directory, e.g. `.wego/profiles/`. Lets add support for specifying a default installation directory, removing the need for the users to repeatedly specify it.

## Implementation
We already have a `.pctl/config.yaml` file that holds the GitRepository resource name/namespace for your Flux repository. This is created during the `pctl bootstrap` command. Lets add an additional flag `pctl bootstrap --default-dir` for specifying the default directory profile installations should go into. Now whenever a user runs `pctl add` and doesn't specify an output directory we will check the `.pctl/config.yaml` to see if this `default-dir` is set. If it is we use that directory as the output.

## Acceptance
1. cd into my Flux repository
2. run `pctl bootstrap --git-repostory foo/bar --default-dir .wego/profiles`
3. run `pctl add --name nginx my-catalog/weaveworks-nginx`
4. see that the `weaveworks-nginx` profile installation is at `.wego/profiles/weaveworks-nginx`. 
5. ???
6. profit

Skarlso:
Aaaand we're back to `--out`, right? :D

aclevername:
> Aaaand we're back to `--out`, right? :D

What do you mean :thinking: ? `--out` is a flag we support already

Skarlso:
ah, so want a flag to save what `--out` will use permanently?

aclevername:
> ah, so want a flag to save what `--out` will use permanently?

basically yeah! So people can enforce a particular directory structure

PR #313, opened 2021-09-16
Add permanent out folder setting into the config

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/309

### Checklist


cmd/pctl/add.go
```diff
@@ -245,6 +250,28 @@ func getGitRepositoryNamespaceAndName(c *cli.Context) (string, string, error) {
...
+	}
+	if c.IsSet("out") {
+		return c.String("out"), nil
+	}
+	if config != nil && config.DefaultDir != "" {
+		return config.DefaultDir, nil
+	}
+	return defaultOut, nil
+}
+
```
aclevername, line 274: Missing some tests! A lot of logic going down here

image

Skarlso: For the record, you don't have any tests for your git-repository in add either. :D

Skarlso: Ah, you do, but in e2e...........

Skarlso: I was about to write a whole thing in `add` for this. :D It didn't occur to me to look in e2e 🤦

Skarlso: Also, that's still Eddy Murphy to me. :D

aclevername: > Ah, you do, but in e2e...........

:smirk: 

but for real yeah unit testing this file SUCKS. So a lot of the tests end up integration, which SUCKS even more. Add unit tests if you can :smile:

aclevername: Also `do as i say not as i do` applies here :joy:

cmd/pctl/add.go
```diff
@@ -245,6 +250,25 @@ func getGitRepositoryNamespaceAndName(c *cli.Context) (string, string, error) {
 	return "", "", fmt.Errorf("flux git repository not provided, please provide the --git-repository flag or use the pctl bootstrap functionality")
 }
 
+// getOutFolder returns the output folder with the following precedence:
+// User set --out overrides local configuration.
+// Local configuration, if set.
+// Default out which is `.`.
+func getOutFolder(c *cli.Context) (string, error) {
```
aclevername, line 257:
```suggestion
func getProfileOutputDirectory(c *cli.Context) (string, error) {
```

I think we try to use `dir/directory` instead of `folder` in general. Also think `out` is a bit vague, `outputDirectory`? or `profileInstallationDirectory`, something a bit more descriptive might be nice

cmd/pctl/add.go
```diff
@@ -245,6 +250,25 @@ func getGitRepositoryNamespaceAndName(c *cli.Context) (string, string, error) {
...
+	if c.IsSet("out") {
+		return c.String("out"), nil
+	}
+	wd, err := os.Getwd()
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch current working directory: %w", err)
+	}
+	config, err := bootstrap.GetConfig(wd)
```
aclevername, line 265: I think we do this already somewhere for getting the `gitRepository` value? might be able to pass it in rather than re-fetching it

cmd/pctl/add.go
Skarlso: Fair point

cmd/pctl/add_test.go
```diff
@@ -0,0 +1,77 @@
...
+				f := &flag.FlagSet{}
+				f.String("out", "user-defined", "")
+				c := cli.NewContext(&cli.App{
+					Commands: []*cli.Command{add},
+				}, f, nil)
+				out, err := getOutFolder(c)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(out).To(Equal("config-dir"))
+			})
+		})
```
aclevername, line 75: missing a test to say "when config file is set and the user has configured it, the users choice takes preference" :smile:

cmd/pctl/add_test.go
```diff
@@ -0,0 +1,77 @@
...
+			f.String("out", "user-defined", "")
+			_ = f.Set("out", "user-defined")
+			c := cli.NewContext(&cli.App{
+				Commands: []*cli.Command{add},
+			}, f, nil)
+			out, err := getOutFolder(c)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(Equal("user-defined"))
+		})
+		When("the user didn't set anything", func() {
```
aclevername, line 40:
```suggestion
		When("the user didn't provide a value or use a config file", func() {
```

cmd/pctl/add.go
Skarlso: We do, yeah. I was wondering about that, but in the end, I wanted to keep its integrity to itself? Like a single, self-sustained thing? But I'll do the refactor and see how it feels. :)

cmd/pctl/bootstrap.go
```diff
@@ -6,21 +6,28 @@ import (
...
 		Flags: []cli.Flag{
 			&cli.StringFlag{
 				Name:  "git-repository",
 				Value: "",
 				Usage: "The namespace and name of the GitRepository object governing the flux repo.",
 			},
+			&cli.StringFlag{
+				Name:  "default-dir",
+				Value: "",
+				Usage: "Directory to always use with pctl add.",
```
aclevername, line 29:
```suggestion
				Usage: "Default directory to place profile installations in",
```

cmd/pctl/add_test.go
Skarlso: huh? I thought I wrote that one...

pkg/bootstrap/bootstrap_test.go
```diff
@@ -39,7 +40,36 @@ var _ = Describe("Bootstrap", func() {
...
+			output, err := cmd.CombinedOutput()
+			Expect(err).ToNot(HaveOccurred(), fmt.Sprintf("init failed: %s", string(output)))
+
+			Expect(bootstrap.CreateConfig(bootstrap.Config{
+				GitRepository: profilesv1.GitRepository{
+					Name:      "bar",
+					Namespace: "foo",
+				},
+				DefaultDir: "default-dir",
+			}, temp)).To(Succeed())
```
aclevername, line 72: does this need to be a separate test? I feel like its an extension of the above test

cmd/pctl/add_test.go
Skarlso: ah shit. :D It's on my todo, but I didn't write it because it wasn't working. I wanted to finish that today. :D

pkg/bootstrap/bootstrap_test.go
Skarlso: It doesn't have to, no.

pkg/bootstrap/bootstrap_test.go
```diff
@@ -50,12 +80,18 @@ var _ = Describe("Bootstrap", func() {
...
 
 		When("the directory is not a git directory", func() {
 			It("returns an error", func() {
-				err := bootstrap.CreateConfig("foo", "bar", temp)
+				err := bootstrap.CreateConfig(bootstrap.Config{
+					GitRepository: profilesv1.GitRepository{
+						Name:      "bar",
+						Namespace: "foo",
+					},
+				}, temp)
```
aclevername, line 94: we create this `bootstrap.Config` a lot, worth creating once in a `BeforeEach` and saving the value?

Skarlso: Yep

README.md
```diff
@@ -308,6 +309,13 @@ These are as follows:
 
 The catalog service options can be configured via `--catalog-service-name`, `--catalog-service-port` and `--catalog-service-namespace`
 
+### Bootstrap
+
+It's possible to bootstrap some values for pctl to use on later calls. To do that, call `pctl bootstrap <values>`.
+
+Bootstrap will create a config file under the set folder with the set values. To find out what values can be configured
+with bootstrap consult the `help` page.
+
```
aclevername, line 318:
```suggestion
### Bootstrap

Some configuration provided during a `pctl add` might be repeated for every command subsequent `pctl add`. To save this information into a config file you can use `pctl bootstrap`. This creates a config file in your git repository, then every  `pctl add` command will check this file and use the values set by default. The user can still specifically provide these values in the comand and they will take president. Currently you can configure:
- `--git-repository` value
- `--out` value

```

something like that?

Skarlso: Yep! Thanks! :)

cmd/pctl/add.go
Skarlso: I know now why it's separate. Because if it's set, we bail out early, not even trying to attempt to do an io operation. It's a small optimisation basically. Not saying that it's worth it, but that was why I didn't extract it. Because we have a path in which it's not used at all.

aclevername: :thinking: you mean if `out` and `git-repository` are set we don't need to read this, so the logic is separate for each call?

Going along with our chat in slack maybe always reading it, and then logging:
1. if we read it/find it
2. if we fail (for any reason)
3. what values we ignore/use

Skarlso: No, in both cases if the user set something, we never read the config file. That's what I meant.

Skarlso: I mean, I can still do the refactor of course. Just wanted to share why I didn't do it. :D Because there is a small chance that we never HAVE to read the file. :)

Skarlso: Done

aclevername (approved):

Issue #260, opened 2021-08-05
pctl upgrade should default to latest version if no version is provided


aclevername:
**What feature/behavior/change do you want?**

**Why do you want this feature?**

Callisto13:
Thanks @sympatheticmoose 👍  My comment was a bit blunt, a quick reminder from a planning discussion 🙈 

> I would think we would want to encourage adoption of the latest updates

Yes definitely agree with that.

> I would expect most upgrade commands to grab latest unless I specify a version as a parameter.

There is a difference between _latest_ and _latest compatible_ version. In the case of upgrading, eg, a CLI tool, then upgrading to latest is nbd. But on a live system, or live-adjacent in the case of profiles, things can get messy and we would be relying on the goodwill of authors to ensure things can be moved forwards without conflicts. At that point we have to decide where the cut off would be.

Our main concern is when, for example, someone has `0.5.0` installed. They call `upgrade` and latest is `0.8.0`. That is quite a significant jump. In many situations like this you would say that they could only jump, say, 1 or 2 minor versions, or go through some patches first. (We could be smart and figure that out for them of course, but that would perhaps assume a rule of compatibility applied to all profiles which authors have to follow).

Obviously the risk for us is way lower, since we don't deploy anything live, so a botched upgrade is annoying more than anything else.

Our initial thought was to have a `--latest` flag or `latest` arg, so if users wanted to yolo it they could.

Himangini:
@sympatheticmoose have we decided on this? shall we close this issue or proceed to implement the suggestion
> Our initial thought was to have a --latest flag or latest arg, so if users wanted to yolo it they could.
and then we can just learn/refine as we get more adoption?

sympatheticmoose:
Up to y'all. I was happy with the suggested plan as stated :)

aclevername:
Out in `0.9.0` @sympatheticmoose :smile:

PR #315, opened 2021-09-17
Add upgrading to the latest version

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/260

### TODO

- [x] Find a nice way to integration test this thing

### Checklist


cmd/pctl/upgrade.go
```diff
@@ -18,8 +19,12 @@ func upgradeCmd() *cli.Command {
 	return &cli.Command{
 		Name:      "upgrade",
 		Usage:     "upgrade profile installation",
-		UsageText: "To upgrade an installation: pctl upgrade pctl-profile-installation-path/ v0.1.1 ",
-		Flags:     createPRFlags,
+		UsageText: "To upgrade an installation: pctl upgrade pctl-profile-installation-path/ v0.1.1",
+		Flags: append(createPRFlags, &cli.BoolFlag{
+			Name:        "latest",
+			Usage:       "--latest",
+			DefaultText: "*WARNING*: Upgrade to the latest version.",
```
aclevername, line 26:
```suggestion
			DefaultText: "*WARNING*: Upgrade to the latest version. Please ensure this is a valid upgrade path before proceeding",
```

cmd/pctl/upgrade.go
```diff
@@ -36,12 +41,24 @@ func upgradeCmd() *cli.Command {
...
+		if c.Args().Len() != 1 {
+			return fmt.Errorf("please provid the path to the profile to upgrade to, e.g. pctl upgrade my-profile/")
+		}
+		profilePath = c.Args().Slice()[0]
+	} else {
+		if c.Args().Len() != 2 {
+			return fmt.Errorf("please provid the path to the profile and version to upgrade to, e.g. pctl upgrade my-profile/ v0.1.1")
+		}
+		profilePath = c.Args().Slice()[0]
+		profileVersion = c.Args().Slice()[1]
```
aclevername, line 59: what about if `latest` was a valid value, rather than a flag :thinking: ?

Skarlso: I don't really like that. If you see the previous push, I would have to then mock about with senver to compare the generated tags and values. Adds complexity. Also this is more explicit in my mind and more easy to discover as an option. While the <version> option doesn't make me think about , oh let's try writing `latest` as version to see what happens.

cmd/pctl/upgrade.go
Skarlso: I had `; whatever that is.` Yours is definitely more polite hahahahahahhaa.

aclevername (approved):

Issue #316, opened 2021-09-17
Command validation does not provide the missing argument for pctl add


steve-fraser:
**What were you trying to accomplish?**

I am trying to add a profile via the `pctl add` command.
**What happened?**

When I run the command, I get the `✗ argument must be provided`, but it does not provide the argument making it hard to debug why the command is not going through. 

**How to reproduce it?**

```
pctl add --name contour \
  --profile-path . \
  --create-pr \
  --pr-repo steve-fraser/weaveworks-testing-playground
```

**Logs**

```
NAME:
   pctl add - generate a profile installation

USAGE:
   To add from a profile catalog entry: pctl --catalog-url <URL> add --name pctl-profile --namespace default --profile-branch main --config-map configmap-name <CATALOG>/<PROFILE>[/<VERSION>]
   To add directly from a profile repository: pctl add --name pctl-profile --namespace default --profile-branch development --profile-repo-url https://github.com/weaveworks/profiles-examples --profile-path bitnami-nginx

OPTIONS:
   --create-pr                   If given, upgrade will create a PR for the modifications it outputs. (default: false)
   --pr-message value, -m value  The message to use for committing. (default: Push changes to remote)
   --pr-remote value             The remote to push the branch to. (default: origin)
   --pr-base value               The base branch to open a PR against. (default: main)
   --pr-branch value             The branch to create the PR from. Generated if not set.
   --pr-repo value               The repository to open a pr against. Format is: org/repo-name.
   --name value                  The name of the installation.
   --namespace value             The namespace to use for generating resources. (default: default)
   --profile-branch value        The branch to use on the repository in which the profile is. (default: main)
   --config-map value            The name of the ConfigMap which contains values for this profile.
   --out value                   Optional location to create the profile installation folder in. This should be relative to the current working directory. (default: current)
   --profile-repo-url value      Optional value defining the URL of the repository that contains the profile to be added.
   --profile-path value          Value defining the path to a profile when url is provided. (default: <root>)
   --git-repository value        The namespace and name of the GitRepository object governing the flux repo.
   --help, -h                    show help (default: false)
   
✗ argument must be provided
(⎈ |kind-prod:default)➜  weaveworks-testing-playground git:(main) pctl add --name contour \
  --profile-path . \
  --create-pr \
  --pr-repo steve-fraser/weaveworks-testing-playground
NAME:
   pctl add - generate a profile installation

USAGE:
[...]

PR #329, opened 2021-09-27
Update command line parsing error

Skarlso:
### Description

Closes https://github.com/weaveworks/pctl/issues/316

### Checklist

### DIFF
--- a/cmd/pctl/main.go
+++ b/cmd/pctl/main.go
@@ -1,19 +1,21 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 
 	"github.com/urfave/cli/v2"
-	"github.com/weaveworks/pctl/pkg/client"
-	"github.com/weaveworks/pctl/pkg/log"
-	"github.com/weaveworks/pctl/pkg/version"
 	profilesv1 "github.com/weaveworks/profiles/api/v1alpha1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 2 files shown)


cmd/pctl/main.go
```diff
@@ -83,7 +85,7 @@ func globalFlags() []cli.Flag {
 
 func parseArgs(c *cli.Context) (string, *client.Client, error) {
 	if c.Args().Len() < 1 {
-		return "", nil, fmt.Errorf("argument must be provided")
+		return "", nil, errors.New("<CATALOG>/<PROFILE>[/<VERSION>] must be provided")
```
Himangini, line 88: can we add a test for this?
also doesn't this affect other commands? I thought `argument must be provided` was thrown by all commands when the required argument was missing 🤔

Skarlso: I can add an integration test.

Skarlso: ah, and to answer your question, nope, we don't have this line anywhere in the docs. :D

Himangini: `parseArgs` is used by `get` command too 🙈

Himangini: oh wait we will never get to `if c.Args().Len() < 1 {` condition in the `get` command 👍🏻

Skarlso: yes, but that part is already tested. it doesn't get into the args part because there is a different check for that in the Get command already

Himangini (approved):
LGTM 👍🏻

